PR #20, opened 2015-09-14
MOTECH-1732: Improves datanucleus performance turning off detachAllOn…

pgesek:
…Commit property

I turned off the datanucleus property detachAllOnCommit. I added the @Transactional
annotations in modules from main repository. Now after this change it will be
problem with update operation. Previously we could update the instance like
the following :
Animal animal = new Animal();
animal = dataService.create(animal);
animal.setYears(5);
dataService.update(animal);

Now we will have the exception because object animal is in the transient state
in update method. It means that instead of updating previous object, datanucleus
will create new one. After this changes if you want to update the object you
should retrieve and update the object in the same transaction. For example using
doInTransaction method from mds services :
Animal animal = new Animal();
dataService.create(animal);

dataService.doInTransaction(new TransactionCallbackWithoutResult() {
	protected void doInTransactionWithoutResult(TransactionStatus status) {
		Animal animalToUpdate = dataService.retrieveAll().get(0);
		animalToUpdate.setYears(5);
		dataService.update(animalToUpdate);
	}
});

I had to turn off this datanucleus property, because of performance issues.
After update datanucleus to version 4, there is a lot of problems with CRUD
operations execution times. This commit should solve the issues.

Conflicts:
	modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java

Change-Id: Ib94c52877a60dd7bf09ee8504e0c54109f5efad2


kmadej:
I've closed the pull request, because it is duplicated with https://github.com/motech/motech/pull/32

PR #27, opened 2015-09-14
MOTECH-1876 Combobox multi-select for defaultValue does not display v…

wstrzelczyk:
…alues

*Fixed problem with displaying values in Combobox for default Value of field.
*Fixed filtering by combobox values.
*Added validation for enum type values if has reserved words.
*Hidden placeholder input field for field type Collection and Boolean.
*Fixed minor issues on UI

### DIFF
--- a/platform/mds/mds-web/src/main/resources/webapp/js/directives.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/directives.js
@@ -1969,13 +1969,12 @@
             restrict: 'A',
             require : 'ngModel',
             link: function (scope, element, attrs, ngModel) {
-                var viewScope = findCurrentScope(scope, 'draft'),
+                var entity, value, resetDefaultValue, checkIfNeedReset,
+                viewScope = findCurrentScope(scope, 'draft'),
                 fieldPath = attrs.mdsPath,
                 fieldId = attrs.mdsFieldId,
-                typeField = attrs.defaultMultiselectList,
-                entity,
-                value,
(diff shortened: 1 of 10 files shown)


platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2570,6 +2581,134 @@
...
+            'while'
+        ];
+        return {
+            require: 'ngModel',
+            link: function(scope, element, attrs, ctrl) {
+                var validateReservedWords,
+                    result = false,
+                    checkReservedWords = function (inputValue) {
+                        result = false;
+                        $.each(RESERVED_WORDS, function (index, reservedWord) {
```
pgesek: use $.inArray(value, array) instead of doing each and checking manually

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2570,6 +2581,134 @@
...
+                    if (newValue !== oldValue) {
+                        ctrl.$setViewValue(ctrl.$viewValue);
+                    }
+                });
+            }
+        };
+    });
+
+    directives.directive('colonValidity', function() {
+        var COLON_REGEXP = /[:]|[;]/;
```
pgesek: what are we validating here, just colons? What's the point, there are other illegal symbols. We should just check if the input is alphanumeric and not a reserved keyword.

PR #30, opened 2015-09-14
Fix MOTECH-1855

jefeweisen:
Fixes: https://applab.atlassian.net/browse/MOTECH-1855


platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -306,6 +306,8 @@ private void removeFromListManyToMany() {
         }
         productDataService.deleteAll();
         supplierDataService.deleteAll();
+        entryDataService.deleteAll();
+        messageDataService.deleteAll();
```
pgesek: 1) Don't these two lines belong in the clearDb() method, rather than removeFromListManyToMany() ?
2) messageDataService.deleteAll() is already called in clearDb() on line 282, is this call still needed?

pgesek:
Closing as this is stale for over a month

PR #32, opened 2015-09-15
MOTECH-1732: Improves datanucleus performance

kmadej:


pgesek:
This PR needs to get updated

platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/DetachedFieldFromTransientObjectException.java
```diff
@@ -0,0 +1,24 @@
...
+    private static final long serialVersionUID = 5232196256542540610L;
+
+    public DetachedFieldFromTransientObjectException() {
+        super("When you want to get detached field from object, you have to retrieve the object from database in the same transaction," +
+                "for example : " +
+                "Object blobValue = dataService.doInTransaction(new TransactionCallback() {\n" +
+                "   @Override\n" +
+                "   public Object doInTransaction(TransactionStatus transactionStatus) {\n" +
+                "       Animal animal = dataService.findById(1L);\n" +
+                "       return dataService.getDetachedField(animal, \"someBlob\");" +
```
pgesek: Why would I do this, can't I just do animal.getSomeBlob() here?

platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/UpdateFromTransientException.java
```diff
@@ -0,0 +1,25 @@
...
+ * Exception, that signalizes updating object from transient state
+ */
+public class UpdateFromTransientException extends MdsException {
+
+    private static final long serialVersionUID = 3815424623777042327L;
+
+    public UpdateFromTransientException() {
+        super("When you want to update object you have to retrieve the object from database in the same transaction," +
+                "for example : " +
+                "dataService.doInTransaction(new TransactionCallbackWithoutResult() {\n" +
```
pgesek: We don't want this doInTransaction() stuff plastered across our error messages. Spring transactions is the preferred way to go

platform/mds/mds/src/main/java/org/motechproject/mds/repository/MotechDataRepository.java
```diff
@@ -167,6 +167,10 @@ public long delete(String[] properties, Object[] values, InstanceSecurityRestric
         return QueryExecutor.executeDelete(query, values, restriction);
     }
 
+    public T getDetachedObject(T object) {
```
pgesek: why not just name this detachCopy() ?

platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java
```diff
@@ -175,6 +179,9 @@ public T retrieve(String primaryKeyName, Object value) {
     @Transactional
     public T update(final T object) {
         validateCredentials(object);
+        if (JDOHelper.getObjectState(object) == ObjectState.TRANSIENT) {
```
pgesek: What if, if in this case, we would try to fetch from the database by ID, then use PropertyUtil to copy properties? If id is not set, or the object not found then throw the exception?

platform/mds/mds/src/main/java/org/motechproject/mds/service/MotechDataService.java
```diff
@@ -138,6 +138,23 @@
     long count();
 
     /**
+     * Returns detached copy of the given object
+     *
+     * @param object the object to be detached
+     * @return the detached copy of the given object
+     */
+    T getDetachedObject(T object);
```
pgesek: Same as before, this is actually getDetachedCopy

platform/mds/mds/src/main/resources/datanucleus.properties
```diff
@@ -13,8 +13,9 @@ datanucleus.schema.validateConstraints=false
 datanucleus.validation.mode=auto
 datanucleus.identifier.case=MixedCase
 datanucleus.plugin.pluginRegistryClassName=org.datanucleus.plugin.OSGiPluginRegistry
-datanucleus.DetachAllOnCommit=true
 datanucleus.classLoaderResolverName=clr.mds
 datanucleus.query.sql.allowAll=true
+datanucleus.RetainValues=true
+datanucleus.manageRelationshipsChecks=false
```
pgesek: any reason this went to false?

platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java
pgesek: Will this method of updating allow us to get rid of those doIntransaction from ITs? We already have an updateFromTransient method if I recall

kmadej:
I will try to explain how it works now after turning off datanucleus property detachAllOnCommit.

1. Persistent objects outside of transaction will be in TRANSIENT state.
2. I set datanucleus property retainValues to true, which means that in TRANSIENT state we should have access to objects fields
3. However there are some types of fields that can be null. For example: BLOB, Comboboxes. For this fields we should use getDetachedField method.
4. CRUD operations on object in TRANSIENT state are allowed. However there will be a problem with updating bidirectional relationships, because other side will not know about changes.
5. Object with bidirectional relationship field should be updated in transaction.
6. You can detach object using detachCopy method from MDS service. Then object outside of transaction will be in DETACHED state.
7. Using of this object will be the same as before my commit.

platform/email/src/main/java/org/motechproject/email/service/impl/EmailSenderServiceImpl.java
```diff
@@ -42,6 +43,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSenderServiceImpl.class);
 
     @Override
+    @Transactional
```
sebbrudzinski: In case we are unable to send mail, we throw an exception from this method, that will rollback the transaction and remove the added log.

sebbrudzinski:
One more thing... I was getting exceptions while attempting to edit CMS Lite String contents. I can see that it doesn't happen without these changes. Can you check that as well?

kmadej:
There is a problem in CMS Lite module. I will push a small fix for service in this module in another PR for modules repo.

kmadej:
test this please

kmadej:
test this please

sebbrudzinski:
@kmadej @pgesek I couldn't find any more problems with those changes

platform/email/src/main/java/org/motechproject/email/service/impl/EmailSenderServiceImpl.java
```diff
@@ -42,6 +43,7 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailSenderServiceImpl.class);
 
     @Override
+    @Transactional(noRollbackFor = MailException.class)
```
pgesek: Will this work when we are participating in an existing transaction? Regardless, I would make MailException a checked exception in this case.

platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -126,6 +127,7 @@ public void save(BootstrapConfig bootstrapConfig) {
     }
 
     @Override
+    @Transactional
```
pgesek: Won't this cause a settings update, since we are operating on the object?

platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -180,6 +185,7 @@ public void evictMotechSettingsCache() {
...
```
pgesek: Do we need a transaction here? Seems its just file operations

platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -661,7 +683,8 @@ public void setConfigAnnotation(Properties configAnnotation) {
      * @param records the records to be searched
      * @return list of file names
      */
-    List<String> getFileNameList(List<ModulePropertiesRecord> records) {
+    @Transactional
```
pgesek: don't see much reason in this being transactional

pgesek:
Transactions will be merged separately from the detaching changes

PR #35, opened 2015-09-16
MOTECH-1860 Bumped external dependencies tag to r030

sebbrudzinski:
The new external osgi bundles release contains OSGified
POI OOXML Schemas bundle, that is required to make POI
library work in OSGi environment.


sebbrudzinski:
test this please

PR #36, opened 2015-09-16
MOTECH-1876 Combobox multi-select for defaultValue does not display v…

wstrzelczyk:
…alues

*Fixed problem with displaying values in Combobox for default Value of field.
*Fixed filtering by combobox values.
*Added validation for enum type values if has reserved words.
*Hidden placeholder input field for field type Collection and Boolean.
*Fixed minor issues on UI in MDS


kmadej:
test this please

PR #37, opened 2015-09-17
MOTECH-1772 QUERY - Server local time is static

pgesek:
Cherry pick of this commit to the 0.26.X branch


pgesek:
test this please

PR #39, opened 2015-09-17
MOTECH-1861: Prevent mds-web from timing out

pgesek:
Mds web was treated as an mds bundle, so it would stop
concurrently with mds. From now on it is treated as a regular
platform bundle.


pgesek:
test this please

PR #40, opened 2015-09-18
MOTECH-1882 Local and global "edit schema" buttons go to the same place

smalecki:


pgesek:
The button navigates correctly to the schema editor correctly, but more than note I have noticed that the list of fields for the entity is not loaded after using the button.

This seems to be the case only in firefox, not in chrome.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2854,26 +2859,29 @@
...
-                });
-
-                unblockUI();
+                if (workInProgress.actualEntity !== $scope.selectedEntity.id) {
+                    blockUI();
+                    workInProgress.setActualEntity(Entities, $scope.selectedEntity.id);
+
+                    $scope.fields = Entities.getFields({id: $scope.selectedEntity.id}, function () {
+                        setSecuritySettings();
+                        setAdvancedSettings();
```
pgesek: seems that after these changes, the abandon changes button does not work properly

smalecki:
test this please

PR #47, opened 2015-09-21
MOTECH-1901: Changed way of embedding MDS grids

ngraczewski:


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -113,7 +113,48 @@
         },
         loadEntity;
 
-    controllers.controller('MdsBasicCtrl', function ($scope, $location, $route, Entities, MDSUtils) {
+    controllers.controller('MdsEmbeddedCtrl', function ($scope, MDSUtils) {
```
pgesek: Is this controller still used in MDS? If yes leave the name, eventually rename it to MdsEmbeddableCtrl

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -3213,4 +3213,23 @@
             }
         };
     });
+
+    directives.directive('embeddedMdsGrid', function($http, $templateCache, $compile) {
```
pgesek: This is not really an embedded mds grid. Its the layout of the page, that might contain the embedded MDS grid if the I happen to choose the right tab. Rename this to something along the lines of TabLayoutWithMdsGrid or something.

pgesek:
We should also make sure that @UIFilterable is working correctly.

PR #49, opened 2015-09-22
MOTECH-1790: Fixes installing a module via the UI

kmadej:
The problem occurs when you work on previous MOTECH version (not the latest one).
While installing a module via the UI, MOTECH used the latest dependency version.
I added methods which should correctly parse properties in pom and its parents.
Thanks to that, properties like for example ${motech.version} will be properly
replaced with for example 0-27-SNAPSHOT.


modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/ModuleAdminServiceImpl.java
```diff
@@ -244,15 +249,25 @@ public ExtendedBundleInformation getBundleDetails(long bundleId) {
         RemoteRepository remoteRepository = new RemoteRepository("central", "default", "http://nexus.motechproject.org/content/repositories/public");
 
         CollectRequest collectRequest = new CollectRequest();
-        collectRequest.setRoot(dependency);
-        collectRequest.addRepository(remoteRepository);
 
-        if (remoteRepositories != null) {
-            for (RemoteRepository repository : remoteRepositories) {
-                collectRequest.addRepository(repository);
+        Dependency updatedDependency = dependency;
```
pgesek: Do we need a separate variable given that its always the same instance?

modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/ModuleAdminServiceImpl.java
```diff
@@ -263,6 +278,66 @@ public ExtendedBundleInformation getBundleDetails(long bundleId) {
...
 
+    private String parseDependencyVersion(Dependency dependency, MavenRepositorySystemSession mvnRepository, RepositorySystem system,
+                                          RemoteRepository remoteRepository, PomInformation pomInformation) throws ArtifactResolutionException {
+        String parsedVersion;
+        String version = dependency.getArtifact().getVersion();
+
+        if (StringUtils.isEmpty(version)) {
+            parsedVersion = "[0,)";
+        } else if (version.contains("${")) {
+            String propertyName = StringUtils.remove(version, "${");
```
pgesek: What if i I have ${baseVersion}-r1 as the version? or ${major.x.version}.${minor.x.version} ? There are better ways of parsing such tokens in Strings rather then doing it by hand.

platform/server-api/src/main/java/org/motechproject/server/api/PomInformation.java
```diff
@@ -0,0 +1,162 @@
...
+    private PomInformation parentPomInformation;
+
+    private Parent parent;
+    private List<Dependency> dependencies;
+    private Properties properties;
+    private List<RemoteRepository> repositories;
+
+    public void parsePom(File file) {
+        try {
+            parsePom(new FileInputStream(file));
```
pgesek: unclosed file stream

platform/server-api/src/main/java/org/motechproject/server/api/PomInformation.java
```diff
@@ -0,0 +1,162 @@
...
+                repositories = new LinkedList<>();
+            }
+
+            parent = model.getParent();
+            setPropertiesFromModel(this, model);
+
+            for (org.apache.maven.model.Dependency dependency : model.getDependencies()) {
+                if (!"test".equalsIgnoreCase(dependency.getScope())) {
+                    dependencies.add(new Dependency(new DefaultArtifact(
+                            (dependency.getGroupId().contains("${")) ? parent.getGroupId() : dependency.getGroupId(),
```
pgesek: This is a bit of a stretch, why not just parse it the same way we do with versions?

platform/server-api/src/main/java/org/motechproject/server/api/PomInformation.java
```diff
@@ -0,0 +1,162 @@
...
+                repositories.add(new RemoteRepository(remoteRepository.getId(), "default", remoteRepository.getUrl()));
+            }
+        } catch (Exception e) {
+            LOGGER.error("Error while reading POM file", e);
+        }
+    }
+
+    public void parseParentPom(File file) {
+        try {
+            parseParentPom(new FileInputStream(file));
```
pgesek: always close streams

platform/server-api/src/test/java/org/motechproject/server/api/PomInformationTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+        Dependency dependency = new Dependency(new DefaultArtifact(
+                "org.motechproject",
+                "motech-osgi-platform",
+                "",
+                "jar",
+                "${project.version}"
+        ), JavaScopes.RUNTIME);
+
+        InputStream inputStream = this.getClass().getClassLoader().getResourceAsStream("pom.xml");
```
pgesek: close streams

platform/server-api/src/test/java/org/motechproject/server/api/PomInformationTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void shouldParseParentPomFile() {
+        Properties properties = new Properties();
+        properties.put("test.properties", "testParent");
+        // Because we use <version> tag in our tested pom, the parsing method should add this as property
+        properties.put("project.version", "0-27-SNAPSHOT");
+
+        InputStream inputStream = this.getClass().getClassLoader().getResourceAsStream("parentPom.xml");
```
pgesek: close streams

modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/ModuleAdminServiceImpl.java
```diff
@@ -263,6 +278,66 @@ public ExtendedBundleInformation getBundleDetails(long bundleId) {
...
+        } else {
+            parsedVersion = version;
+        }
+
+        return parsedVersion;
+    }
+
+    private String getVersionFromProperties(String propertyName, MavenRepositorySystemSession mvnRepository, RepositorySystem system,
+                                            RemoteRepository remoteRepository, PomInformation pomInformation) throws ArtifactResolutionException {
+        Properties properties = pomInformation.getProperties();
```
pgesek: Will using ${project.version} as the dependency version work?

platform/server-api/src/main/java/org/motechproject/server/api/PomInformation.java
```diff
@@ -0,0 +1,162 @@
...
+     * version tag is used in the pom file.
+     *
+     * @param pomInformation the information about parsed pom file
+     * @param model the model from parsed pom file
+     */
+    private void setPropertiesFromModel(PomInformation pomInformation, Model model) {
+        Properties propertiesFromModel = model.getProperties();
+
+        if (model.getVersion() != null) {
+            propertiesFromModel.put("project.version", model.getVersion());
```
pgesek: Regarding my earlier question, I see that it will work for ${project.version} since we are hardcoding it. I am assuming other maven variables will not work however?

modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/ModuleAdminServiceImpl.java
```diff
@@ -263,6 +278,66 @@ public ExtendedBundleInformation getBundleDetails(long bundleId) {
...
+
+        if (StringUtils.isEmpty(version)) {
+            parsedVersion = "[0,)";
+        } else if (version.contains("${")) {
+            String propertyName = StringUtils.remove(version, "${");
+            propertyName = StringUtils.remove(propertyName, '}');
+
+            parsedVersion = getVersionFromProperties(propertyName, mvnRepository, system, remoteRepository, pomInformation);
+            if (parsedVersion == null) {
+                throw new MotechException(String.format("The property %s used in dependency %s cannot be found in pom " +
```
pgesek: Is it wise for use to throw an exception, given that our parsing seems limited? Perhaps just log an error and try to proceed anyway?

PR #53, opened 2015-09-23
MOTECH-1865: Added unique & required constraints to the domain

pgesek:
We were missing these on some classes, leaving the integrity
ultimately to the application level, which was far from perfect.

Conflicts:
	modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/ActionParameter.java


kmadej:
It looks good to me

PR #57, opened 2015-09-23
MOTECH-1898: Changes the default value for columns to export.

kmadej:
Now the default value is 'currently selected'. Regardless of what user selected
, all the time the default value will be 'currently selected' (also after log out,
log in).


sebbrudzinski:
@pgesek OK to merge

PR #60, opened 2015-09-24
MOTECH-1907 DEV - BUG - Reports - Today button on LocalDate date-pick…

wstrzelczyk:
Fix problem with setting date when click button 'today'.


pgesek:
We should update all datepickers and time pickers across Motech.

pgesek:
test this please

kmadej:
It works as expected in mds-web, however in scheduler and email there is a problem. When we choose 'now', value in the input is correct, however value 'Time' in date time picker is wrong and after clicking 'Done' the wrong (not actual) time is put in the input.

PR #64, opened 2015-09-25
MOTECH-1915: Map field in MDS embedded UI

tstalka:


pgesek:
I've set CallDetailRecord as editable and I have two problems while editing in the embedded view:
1) Clicking new key/value pair does nothing
2) The combobox input has no options

PR #65, opened 2015-09-25
MOTECH-1901: Fixed @UIFilterable annotation and embedded MDS grids

ngraczewski:


pgesek:
We are not supporting Java 8 classes in 0.27

PR #69, opened 2015-09-28
Motech-1859

LukSkarDev:
MdsBundleIT now run twice - with and without l2 cache


platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -168,14 +183,45 @@ public void tearDown() throws Exception {
 
     @Test
     public void testEntitiesBundleInstallsProperly() throws Exception {
+
+        //Run tests with l2 cache active
+        testEntitiesBundleInstallsProperly(true);
+
+        //Turn off l2 cache
+        outputStream =  new FileOutputStream(configurationFile);
```
pgesek: Always close streams - you can use try with resource. Moreover I don't see a point the stream being stored as an instance variable.

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -168,14 +183,45 @@ public void tearDown() throws Exception {
...
+        datanucleusProps.store(outputStream, null);
+        outputStream.close();
+
+        //Run tests without l2 cache
+        testEntitiesBundleInstallsProperly(false);
+
+        //Remove added l2 cache property after test was done
+        //New FileOutputStream is created so the config won't be appended
+        //to existing one
+        outputStream = new FileOutputStream(configurationFile);
```
pgesek: as above, must always be closed

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -168,14 +183,45 @@ public void tearDown() throws Exception {
...
+        datanucleusProps.remove("datanucleus.cache.level2.type", "none");
```
pgesek: what if it was none to being with? This is academical, but would be better to store the value of the original property. Moreover this whole operation should be performed in teardown, currently it will only revert the setting if the test succeeds

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -168,14 +183,45 @@ public void tearDown() throws Exception {
 
     @Test
     public void testEntitiesBundleInstallsProperly() throws Exception {
```
pgesek: Would be best to create a separate method for changing the DataNucleus property

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -137,6 +142,8 @@
     private JarGeneratorService generator;
     private EntityService entityService;
     private MotechDataService service;
+    private FileOutputStream outputStream;
```
pgesek: this should be a local variable

pgesek:
Please provided better descriptions of pull requests

PR #71, opened 2015-09-28
MOTECH-1844 Fixed MDS Data browser for lazy loaded fields

sebbrudzinski:
This commit fixes issues viewing instances that have
lazy loaded fields. The Spring controller attempted to
access these fields during serialization to json, what
resulted in JDO exception, due to accessing lazy loaded
fields outside a transaction. The relationships now
have their related fields set to null for display.

### DIFF
--- a/platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
+++ b/platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
@@ -77,7 +77,6 @@ import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -442,7 +441,7 @@ public class InstanceServiceImpl implements InstanceService {
                 throw new ObjectNotFoundException(service.getClassType().getName(), instanceId);
             }
             fieldRecord = new FieldRecord(field);
-            fieldRecord.setValue(parseValueForDisplay(instance, field.getMetadata(Constants.MetadataKeys.RELATED_FIELD)));
+            fieldRecord.setValue(parseValueForDisplay(instance, field.getMetadata(Constants.MetadataKeys.RELATED_CLASS)));
             fieldRecord.setDisplayValue(instance.toString());
             return fieldRecord;
         } catch (IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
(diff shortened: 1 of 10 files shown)


sebbrudzinski:
Test this please

sebbrudzinski:
test this please

sebbrudzinski:
Please take an extra look while testing, if browsing relationships does not remove them. There was a known issue that caused relationships to dissapear completely, if removal was called from transactions, while we only want to remove them for UI purposes.

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -952,25 +953,32 @@ private Object parseValueForDisplay(Object value, MetadataDto relatedFieldMetada
...
         }
 
         return parsedValue;
     }
 
-    private Object removeCircularRelations(Object object, String relatedField) {
-        // we must also handle a field that is a collection
-        // because of this we handle regular fields as single objects collection here
-        Collection objectsCollection = (object instanceof Collection) ? (Collection) object : Arrays.asList(object);
+    private Object breakDeepRelationChainForDisplay(Object value, String relatedClassName) {
```
pgesek: We need some sort of comment discalimer about not using transactions on methods that use this.

pgesek: Hm maybe that would be too many methods

pgesek: Ok, here's what we should do. We should either try using @Transactional(propagation=NEVER), but that would have to go on the methods that use this, since this is not a public method. Eventually we can try manually checking the transaction manager here, and throw an exception if a transaction exists. I'm wary of leaving it like this to be honest.

pgesek: That being said, the best way to handle this would be to move the method to a separate helper component. That way you can use the Transactional annotation with propagation=NEVER on this method. Generally speaking we should probably be moving out parts of the logic from InstanceServiceImpl, in order to not bloat it too much.

pgesek: Would also be good to add a unit test for the said component.

pgesek:
This seems to work fine. You only need to incorporate my comment about using a separate component with Transaction(propagation=never) to prevent future issues. And add a unit test for that component.

PR #74, opened 2015-09-28
 MOTECH-1915: Map field in MDS embedded UI

tstalka:
This commits repairs invisible map inputs. Now user can add/delete rows in the map field and can
see options in the combobox list.


pgesek:
test this please

PR #78, opened 2015-09-28
Removed checks for Tomcat processes

pgesek:
I am removing this for now, since with multiple builds running
this can incorectly stop the server or read it as running.
If we have leftover Tomcat processes, we will get to know about it
sooner or later on the CI. In the future we might consider
comming up with a more sophisticated check.


pgesek:
test this please

pgesek:
Going to close this for now

PR #79, opened 2015-09-28
Motech-1673 Improve UI representation of Maps in Tasks Module

smalecki:


ngraczewski:
Dragging beans onto the text areas doesn't work. Also I would add some minimal/maximal sizes for the text areas and style them similar to the other text boxes.

PR #86, opened 2015-09-30
MOTECH-1806: Errors pointing to logs when a bean validation fails

shubhambeehyv:
when creating an instance through the UI (and the log error isn't helpful either)

Added message from ConstraintViolation to Exception message.

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/ObjectCreateException.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/ObjectCreateException.java
@@ -2,6 +2,10 @@ package org.motechproject.mds.ex.object;
 
 import org.motechproject.mds.ex.MdsException;
 
+import javax.validation.ConstraintViolation;
+import javax.validation.ConstraintViolationException;
+import java.util.Set;
+
 /**
  * Signals that it was not possible to update object instance from the provided data.
  */
@@ -10,6 +14,17 @@ public class ObjectCreateException extends MdsException {
     private static final long serialVersionUID = -6214111291407582493L;
 
     public ObjectCreateException(String entityName, Throwable cause) {
-        super("Unable to create of entity " + entityName, cause, "mds.error.objectUpdateError");
+        super("Unable to create of entity " + entityName + ". " + getMessageFromCause(cause), cause, "mds.error.objectUpdateError");
+    }
+
+    private static String getMessageFromCause(Throwable cause) {
+        String message = "";
+        if (cause instanceof ConstraintViolationException) {
+            Set<ConstraintViolation<?>> violations = ((ConstraintViolationException) cause).getConstraintViolations();
+            for (ConstraintViolation violation : violations) {
+                message += String.format("Field %s %s\n", violation.getPropertyPath(), violation.getMessage() );
+            }
+        }
+        return message;
     }
 }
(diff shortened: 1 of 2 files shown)


platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/ObjectCreateException.java
```diff
@@ -10,6 +14,17 @@
     private static final long serialVersionUID = -6214111291407582493L;
 
     public ObjectCreateException(String entityName, Throwable cause) {
-        super("Unable to create of entity " + entityName, cause, "mds.error.objectUpdateError");
+        super("Unable to create of entity " + entityName + "." + getMessageFromCause(cause), cause, "mds.error.objectUpdateError");
```
kmadej: ". " - add space here

platform/mds/mds/src/main/java/org/motechproject/mds/ex/object/ObjectCreateException.java
```diff
@@ -10,6 +14,17 @@
...
+    }
+
+    private  static String getMessageFromCause(Throwable cause) {
```
kmadej: private static - remove additional space here

kmadej:
It works as expected for creating instance. However there is a problem with updating. Method update from mds service in saveInstance method from InstanceServiceImpl class doesn't throw ConstraintViolationException, it means that now we don't catch this exception and don't throw our ObjectUpdateException. 
The ConstraintViolationException is thrown when the transaction in saveInstance method is being committed. This problem has to be solved, because now we cannot properly propagate ConstraintViolationException while updating instance.

shubhambeehyv:
I tried to update an entity and it was throwing ObjectUpdateException with cause ConstraintViolationException

org.motechproject.mds.ex.object.ObjectUpdateException: Unable to object of entity Task with id 1.Field description may not be empty

	at org.motechproject.mds.web.service.impl.InstanceServiceImpl.saveInstance(InstanceServiceImpl.java:165)
	at org.motechproject.mds.web.service.impl.InstanceServiceImpl.saveInstance(InstanceServiceImpl.java:119)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
......

Caused by: javax.validation.ConstraintViolationException: Validation failed for org.motechproject.tasks.domain.Task@640613dd during pre-persist for groups [interface javax.validation.groups.Default] - exceptions are attached
	at org.datanucleus.validation.BeanValidatorHandler.validate(BeanValidatorHandler.java:77)
	at org.datanucleus.validation.BeanValidatorHandler.prePersist(BeanValidatorHandler.java:105)
........

kmadej:
This is strange behaviour. Previously I checked it using EmailRecord entity and the problem I described above occurred, however in Task entity it works correctly. I will dig into this problem and probably create another ticket for this. Please just correct this PR considering my two comments

shubhambeehyv:
I have edited the code.

kmadej:
Needs rebase

shubhambeehyv:
rebase done

PR #89, opened 2015-09-30
MOTECH-1925: Import CSV is not possible for readonly entities

kmadej:
Now the import button is invisible above the entities grid for readonly entities.
Additionally I disallowed it at the backend.


platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/InstanceService.java
```diff
@@ -282,7 +282,15 @@
...
+     *
      * @param entityId the id of the entity
      */
     void verifyEntityAccess(Long entityId);
+
+    /**
+     * Checks whether the entity with the given ID is non editable.
+     *
+     * @param entityId the id of the entity
+     */
```
tstalka: Add @throws

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -103,7 +103,7 @@ public Object saveInstance(EntityRecord entityRecord) {
     public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
         EntityDto entity = getEntity(entityRecord.getEntitySchemaId());
         validateCredentials(entity);
-        validateNonEditableProperty(entity);
+        validateNonEditableProperty(entity.getId());
```
tstalka: Now there will be two queries to the database instead of one query, maybe we should overload this method.

platform/mds/mds-web/src/main/resources/webapp/partials/widgets/entityInstances.html
```diff
@@ -22,7 +22,7 @@ <h4 class="frame-title">{{selectedEntity.module}} <span ui-if="selectedEntity.mo
             </select>
         </span>
         <div class="btn-group">
-            <button ng-click="importEntityInstances()" type="button" class="btn btn-default" ng-show="showImportButton">
+            <button ng-click="importEntityInstances()" type="button" class="btn btn-default" ng-hide="shouldHideButton()">
                 <span class="glyphicon glyphicon-import"></span>
```
tstalka: showImportButton is needed for custom-ui, we can have editable entity but button should be hidden

PR #95, opened 2015-10-01
MOTECH-1931: Made ordering even more consistent

pgesek:
* lookups are order based on their fields default
* moved query param building code to a separate util


pgesek:
test this please

PR #101, opened 2015-10-02
MOTECH-1903: UI will be blocked until advanced entity settings are re…

ngraczewski:
…trieved


pgesek:
The button is still visible dissapearing in the background. Can we change that it's not displayed until the settings are retrieved? Currently it seems to remember the last state, meaning that when I go from an entity with filters to an entity without them, I can observe the button dissapearing in blocked background.

ngraczewski:
test this please

PR #102, opened 2015-10-03
MOTECH-1388: Added listEntitiesByBundle

tstalka:
Now we can retrieve all entities by bundle symbolic name. This change will
improve performance of embedded UI in mTraining module.

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
@@ -810,7 +810,7 @@ public class EntityServiceImpl implements EntityService {
         }
 
         List<EntityDto> entityDtos = new ArrayList<>();
-        for (Entity entity : allEntities.retrieveAll("bundleSymbolicName", bundleSymbolicName)) {
+        for (Entity entity : allEntities.retrieveBySymbolicName(bundleSymbolicName)) {
             if (entity.isActualEntity()) {
                 entityDtos.add(entity.toDto());
             }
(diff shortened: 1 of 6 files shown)


platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -801,6 +802,23 @@ public void deleteEntity(Long entityId) {
...
 
+    @Override
+    @Transactional
+    public List<EntityDto> listEntitiesByBundle(String bundleSymbolicName) {
+        if (StringUtils.isBlank(bundleSymbolicName)) {
+            throw new MdsException("Bundle symbolic name cannot be empty or null");
+        }
+
+        List<EntityDto> entityDtos = new ArrayList<>();
+        for (Entity entity : allEntities.retrieveAll("bundleSymbolicName", bundleSymbolicName)) {
```
pgesek: We should create a method retrieveBySymbolicName() in allEntities that does the retrieval and filtering.

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -801,6 +802,23 @@ public void deleteEntity(Long entityId) {
...
 
+    @Override
+    @Transactional
+    public List<EntityDto> listEntitiesByBundle(String bundleSymbolicName) {
+        if (StringUtils.isBlank(bundleSymbolicName)) {
+            throw new MdsException("Bundle symbolic name cannot be empty or null");
+        }
+
+        List<EntityDto> entityDtos = new ArrayList<>();
+        for (Entity entity : allEntities.retrieveBySymbolicName(bundleSymbolicName)) {
```
pgesek: I meant that this could also go the repo class, since thats what we do in the other methods there.

PR #111, opened 2015-10-06
MOTECH-1782 Allow the scheduler to act as a task trigger

adamkalmus:


pgesek:
Will need to be reimplemented and moved to the scheduler

PR #116, opened 2015-10-12
MOTECH-1649 preview

jefeweisen:
Code not ready for merge, but please review.  Please do not merge yet.

See instructions, known issues, and comments in Jira:

    https://applab.atlassian.net/browse/MOTECH-1649


platform/mds/mds-secondary-test-bundle/src/test/java/org/motechproject/mds/test/secondary/osgi/MdsSecondBundleIT.java
```diff
@@ -18,6 +18,7 @@
 import org.motechproject.mds.test.secondary.service.MessageRecordDataService;
```
jefeweisen: Start here

modules/testJdoDiscriminator/pom.xml
```diff
@@ -0,0 +1,288 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jefeweisen: Sorry for the lengthy presentation of testJdoDiscriminator in the gihub view.  It's not part of the production code but enables the instructions to work as written.

platform/mds/mds-web/src/main/resources/webapp/partials/widgets/entityInstanceFields.html
```diff
@@ -1,30 +1,7 @@
 <form name="instanceFieldsForm" class="form-horizontal inside" ng-init="getHistory()">
     <div class="form-group" ng-repeat="field in fields" ng-class="{ 'has-error' : {{field.name}}.$invalid && {{field.name}}.$dirty}"
          ng-if="shouldShowInputForField(field)">
-        <label class="col-md-2 col-sm-3 control-label">
```
jefeweisen: Extracted to entityInstanceField.html and entityInstanceField in directives.js

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/domain/Discr1.java
```diff
@@ -0,0 +1,17 @@
+package org.motechproject.mds.test.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Discriminated;
+@Entity
```
pgesek: add empty line above Entity

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/domain/Discr1.java
```diff
@@ -0,0 +1,17 @@
...
+@Discriminated
+abstract public class Discr1 {
+    private Long id;
```
pgesek: In the Jira ticket you mentioned that entities need to define the id field. Any reason for that? The id field is always added automatically anyway.

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/domain/Discr1.java
```diff
@@ -0,0 +1,17 @@
+package org.motechproject.mds.test.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Discriminated;
+@Entity
+@Discriminated
+abstract public class Discr1 {
```
pgesek: Can we rename this to DiscrParent? Easy to get lost with names like this

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/domain/Discr1Start.java
```diff
@@ -0,0 +1,47 @@
+package org.motechproject.mds.test.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@Entity
+public class Discr1Start
```
pgesek: I'm already lost, can we rename this class or at least add a comment explaining the structure?

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/service/DiscrMultipleDataService.java
```diff
@@ -0,0 +1,7 @@
+package org.motechproject.mds.test.service;
+
+import org.motechproject.mds.test.domain.Discr1Multiple;
+
+import org.motechproject.mds.service.MotechDataService;
+public interface DiscrMultipleDataService extends MotechDataService<Discr1Multiple> {
```
pgesek: add an empty line below the import

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/service/DiscrOneToOneDataService.java
```diff
@@ -0,0 +1,7 @@
+package org.motechproject.mds.test.service;
+
+import org.motechproject.mds.service.MotechDataService;
+import org.motechproject.mds.test.domain.Discr1OneToOne;
+
+public interface DiscrOneToOneDataService  extends MotechDataService<Discr1OneToOne> {
```
pgesek: double space

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -779,6 +812,130 @@ public void testManyToManyRelationshipListWithCustomTableNames() {
...
+    public Discr1CaseB createCaseB(Discr1CaseB caseB) {
+        discrDataService.create(caseB);
+        discrCaseBDataService.create(caseB);
+        return caseB;
+    }
+
+    private void saveAndLoadDiscrs(int cCaseA,int cCaseB) {
+        Discr1Multiple start1 = discrMultipleDataService.create(new Discr1Multiple());
+
+        for(int i=0; i<cCaseA; i++)
```
pgesek: oh boy

pgesek: requires some sort of explanation

pgesek: either refactor the method to be a bit more clear, or add some comments

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -779,6 +812,130 @@ public void testManyToManyRelationshipListWithCustomTableNames() {
...
+        assertEquals(cCaseAActual, cCaseA);
+        assertEquals(cCaseBActual, cCaseB);
+    }
+
+    @Test
+    public void canStoreDiscrAsOnetoone() {
+        Discr1Start start = discrStartDataService.create(new Discr1Start());
+
+        Discr1CaseA caseA = createCaseA(new Discr1CaseA(142));
+        AddAsOnetoone(start, caseA);
```
pgesek: follow java conventions when naming methods

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/InstanceService.java
```diff
@@ -83,6 +84,10 @@
      */
     List<FieldDto> getEntityFields(Long entityId);
 
+    List<FieldDto> getEntityFieldsGivenInstance(EntityDto entity, Object instance);
```
pgesek: these methods need javadoc

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -123,17 +123,24 @@ public Object saveInstance(EntityRecord entityRecord) {
...
+
+        DecodePseudofields(entityRecord, entityFields, fieldrecords, entityTarget, deleteValueFieldId);
 
-        List<FieldDto> entityFields = getEntityFields(entityRecord.getEntitySchemaId());
+        entityTarget[0].setId(entityRecord.getId()); // TODO: is this redundant?
+
+        return SaveInstanceI(entityTarget[0], deleteValueFieldId, entityFields, fieldrecords);
+    }
 
+    private Object SaveInstanceI(EntityDto entity, Long deleteValueFieldId,
```
pgesek: follow java conventions

pgesek: and remove the I from the end of the method name

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -123,17 +123,24 @@ public Object saveInstance(EntityRecord entityRecord) {
     @Override
     @Transactional
     public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
-        EntityDto entity = getEntity(entityRecord.getEntitySchemaId());
-        validateCredentials(entity);
-        validateNonEditableProperty(entity);
+        EntityDto[] entityTarget = new EntityDto[]{null};
```
pgesek: avoid arrays, prefer collections. Moreover, why do we initiate it with a null value?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -160,12 +167,91 @@ public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
...
+     *
+     * One pseudofield represents the discriminator.
+     * Once a discriminator is selected, the UI will expand the field list using fieldsAdded.
+     *
+     * Encoding of pseudofields is performed by:
+     *      org.motechproject.mds.service.impl.EntityServiceImpl.EncodePseudofields()
+     */
+    private void DecodePseudofields(EntityRecord entityRecord, List<FieldDto> entityFields, List<FieldRecord> fieldrecords, EntityDto[] entityTarget, Long deleteValueFieldId)
+    {
+        EntityDto entity1 = getEntity(entityRecord.getEntitySchemaId());
```
pgesek: maybe use names like parent and child instead of just adding numbers to variables

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -160,12 +167,91 @@ public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
...
+        }
+    }
+
+    private EntityDto DecodeEntityFromPfds(EntityRecord entityRecord, EntityDto entity, Long deleteValueFieldId)
+    {
+        List<FieldRecord> entityFields = entityRecord.getFields();
+
+        FieldRecord fdSubclass = null;
+        for(FieldRecord fd : entityFields) {
+            if("subclass".equals(fd.getName()))
```
pgesek: If this is a new field we add to entities, the name should get placed in some sort of constant

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -160,12 +167,91 @@ public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
...
+        for(EntityDto entityChild : fdSubclass.getEntitiesDerived()) {
+            if(classnameSimple2.equals(entityChild.getName()))
+                classname2 = entityChild.getClassName();
+        }
+        return getEntity(classname2);
+    }
+
+
+    private List<FieldRecord> DecodeFieldlistFromPfds(List<FieldDto> entityFields, EntityRecord entityRecord, EntityDto entity) {
+        // *** Note that this method mutates entityFields ***
```
pgesek: how, it just seems to iterate over them. Or am I missing something?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -160,12 +167,91 @@ public Object saveInstance(EntityRecord entityRecord, Long deleteValueFieldId) {
...
+
+        List<FieldRecord> fieldrecords;
```
pgesek: just declare and define this on one line

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3736,6 +3736,10 @@
             }
         };
 
+        $scope.deriveIsDiscriminated = function (e) {
+            return !!(e.fields.filter(fd => fd.name === 'subclass')[0]);
```
pgesek: what if there is no such field?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3743,13 +3747,34 @@
         */
         $scope.addEntityInstance = function () {
             blockUI();
-
-            var values = $scope.currentRecord.fields;
+            $scope.addEntityInstanceBase($scope.currentRecord);
+        };
+        $scope.addEntityInstanceBase = function (inst)
```
pgesek: add empty line above this

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3743,13 +3747,34 @@
...
+        {
+            var values = inst.fields;
             angular.forEach (values, function(value, key) {
                 value.value = value.value === 'null' ? null : value.value;
             });
+            for(var i in inst.fields) {
+                var fd = inst.fields[i];
+
+                // {readonly, uiFilterable, . . .} = fields(FieldDto) - fields(FieldRecord)
+                delete fd["readOnly"];
```
pgesek: what's the reason for doing the deletes?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3743,13 +3747,34 @@
...
+                delete fd["uiFilterable"];
+                delete fd["uiChanged"];
+                delete fd["lookups"];
+
+                // TODO: where is "discriminated" coming from?  Not isDiscriminated?
+                var entitiesDerived = fd.entitiesDerived;
+                if(entitiesDerived) {
+                    for (var j in entitiesDerived) {
+                        var e = entitiesDerived[j];
```
pgesek: please use better variable names, this is hard to understand

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3743,13 +3747,34 @@
...
+                // {readonly, uiFilterable, . . .} = fields(FieldDto) - fields(FieldRecord)
+                delete fd["readOnly"];
+                delete fd["uiFilterable"];
+                delete fd["uiChanged"];
+                delete fd["lookups"];
+
+                // TODO: where is "discriminated" coming from?  Not isDiscriminated?
+                var entitiesDerived = fd.entitiesDerived;
+                if(entitiesDerived) {
+                    for (var j in entitiesDerived) {
```
pgesek: Generally speaking this construct should be avoided: http://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-such-a-bad-idea

I can see this iterates over indexes for you? I believe that depenends on the browser.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3865,13 +3890,35 @@
...
+                if (!scope.allEntityFields[i].nonDisplayable) {
+                    availableFieldsForDisplay.push(scope.allEntityFields[i]);
+                }
+            }
+            var subclassnameCurrent = scope.$parent.$parent.$parent.subclassCurrent;
+            if(subclassnameCurrent) {
+                var entitiesT = scope.$parent.$parent.$parent.entitiesDerived;
+                var entitySelected = !entitiesT ? null : entitiesT.filter(function(e) {return e.name === subclassnameCurrent;})[0];
+                var fieldsAdded1 = !entitySelected? null : entitySelected.fieldsAdded;
+                var fieldsAdded = fieldsAdded1.slice(6);  // TODO: remove hack
```
pgesek: whats the difference between fieldsAdded an fieldsAdded1?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3865,13 +3890,35 @@
...
-                    $scope.availableFieldsForDisplay.push($scope.allEntityFields[i]);
+            var availableFieldsForDisplay = [];
+            for (i = 0; i < scope.allEntityFields.length; i += 1) {
+                if (!scope.allEntityFields[i].nonDisplayable) {
+                    availableFieldsForDisplay.push(scope.allEntityFields[i]);
+                }
+            }
+            var subclassnameCurrent = scope.$parent.$parent.$parent.subclassCurrent;
+            if(subclassnameCurrent) {
+                var entitiesT = scope.$parent.$parent.$parent.entitiesDerived;
```
pgesek: What does the T stand for?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3865,13 +3890,35 @@
...
-            for (i = 0; i < $scope.allEntityFields.length; i += 1) {
-                if (!$scope.allEntityFields[i].nonDisplayable) {
-                    $scope.availableFieldsForDisplay.push($scope.allEntityFields[i]);
+            var availableFieldsForDisplay = [];
+            for (i = 0; i < scope.allEntityFields.length; i += 1) {
+                if (!scope.allEntityFields[i].nonDisplayable) {
+                    availableFieldsForDisplay.push(scope.allEntityFields[i]);
+                }
+            }
+            var subclassnameCurrent = scope.$parent.$parent.$parent.subclassCurrent;
```
pgesek: What's with the parents? Don't think it's wise to do something like this. If you absolutely must go up the chain, maybe use the findCurrentScope() method that we have defined in directives.js.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3865,13 +3890,35 @@
...
+            if(subclassnameCurrent) {
+                var entitiesT = scope.$parent.$parent.$parent.entitiesDerived;
+                var entitySelected = !entitiesT ? null : entitiesT.filter(function(e) {return e.name === subclassnameCurrent;})[0];
+                var fieldsAdded1 = !entitySelected? null : entitySelected.fieldsAdded;
+                var fieldsAdded = fieldsAdded1.slice(6);  // TODO: remove hack
+                if(fieldsAdded && fieldsAdded.length>0) {
+                    availableFieldsForDisplay = $scope.allEntityFields.concat(fieldsAdded);
```
pgesek: this code looks really suspect with all the hacks and $parent.$parent stuff. Moreover I'm not sure I understand what are we trying to do here. Are we adding fields from the child class to the parent view? Is that it?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3890,7 +3937,7 @@
 
               $http.get('../mds/entities/'+$scope.selectedEntity.id+'/entityFields').success(function (data) {
                    $scope.allEntityFields = data;
-                   $scope.setAvailableFieldsForDisplay();
+                   $scope.setAvailableFieldsForDisplay($scope);
```
pgesek: What's the point of passing scope as var? Are we doing recursion here? I think I missed it. Or is this called from somewhere else with a different scope?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3927,14 +3974,22 @@
                             dbUserPreferences = $scope.getDataBrowserUserPreferencesCookie($scope.selectedEntity);
 
                         $scope.selectedFields = [];
+                       $scope.entitiesDerived = [];
+                       $scope.fieldsBase = [];
+                       var entitiesT = [];
```
pgesek: again, the cryptic T

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2189,6 +2189,47 @@
         };
     });
 
+    directives.directive('entityInstanceFields2', function (Entities) {
```
pgesek: why is this called entityInstanceFields2? Is this a temporary thing?

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2189,6 +2189,47 @@
...
+        return {
+            restrict: 'AE',
+            transclude: true,
+            scope: true,
+            controller: ['$scope', function ($scope) {
+                $scope.subclassCurrent = $scope.$parent.$parent.$parent.subclassCurrent;
```
pgesek: I would avoid this

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2189,6 +2189,47 @@
...
+                        function(valueNew, valueOld) {
+                        if(valueNew) {
+                            $timeout(function() {
+                                // TODO: six parents is poor coupling with the template layouts.
+                                // I need to carefully review the recently written docs for reuse
+                                // of the data browser https://github.com/motech/motech/pull/100
+                                // and see if I can do something stylistically compatible, while not
+                                // giving up my plan to make the data browser be able to recurseively
+                                // display related instances.
+                                $scope.$parent.$parent.$parent.$parent.$parent.$parent.subclassCurrent = valueNew;
```
pgesek: As mentioned, we have a method for finding the scope, maybe use that. But from what I can see we only to do this for the 'draft' method, I don't see us doing anything like this, but I might be missing something. Why do you need to go so far up this chain? Maybe there is a better way to pass these values?

platform/mds/mds/src/main/java/org/motechproject/mds/annotations/Discriminated.java
```diff
@@ -0,0 +1,13 @@
...
+
+import javax.jdo.annotations.IdentityType;
+import javax.jdo.annotations.PersistenceCapable;
+import java.lang.annotation.*;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@PersistenceCapable(identityType = IdentityType.DATASTORE, detachable = "true")
+public @interface Discriminated {
```
pgesek: This needs javadoc first of all. Moreover I am wondering about the purpose of this annotation. Why not use the JDO inheritance annotations? Is the UI behavior the main difference? Am I understanding this change correctly?

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
```diff
@@ -134,8 +139,10 @@ public void addHelperClassMetadata(JDOMetadata jdoMetadata, ClassData classData,
         cmd.setIdentityType(IdentityType.APPLICATION);
         cmd.setPersistenceModifier(ClassPersistenceModifier.PERSISTENCE_CAPABLE);
 
-        InheritanceMetadata imd = cmd.newInheritanceMetadata();
-        imd.setCustomStrategy("complete-table");
+        if(entityType == EntityType.STANDARD)
+            fixDiscriminator(definition, cmd, 1);
+        // TODO: entityType == EntityType.STANDARD: did trash, and history need inheritanceMetadata?
+        // TODO: do trash and history need discriminatorMetadata?
```
pgesek: Don't have an anwser to this since I'm still struggling with the concept, but they should retain all values, so if we are adding fields from the parent to an entity, then they should retain those values.

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
```diff
@@ -134,8 +139,10 @@ public void addHelperClassMetadata(JDOMetadata jdoMetadata, ClassData classData,
         cmd.setIdentityType(IdentityType.APPLICATION);
         cmd.setPersistenceModifier(ClassPersistenceModifier.PERSISTENCE_CAPABLE);
 
-        InheritanceMetadata imd = cmd.newInheritanceMetadata();
-        imd.setCustomStrategy("complete-table");
+        if(entityType == EntityType.STANDARD)
+            fixDiscriminator(definition, cmd, 1);
```
pgesek: In this class, 'fix' is a term we applied to methods that work on the metadata after the enhancer runs. This is executed during regular generation, so I would start the name with 'generate' or something along those lines.

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
```diff
@@ -144,6 +151,43 @@ public void addHelperClassMetadata(JDOMetadata jdoMetadata, ClassData classData,
...
+                discriminated = (Discriminated) annotation;
+
+        if(discriminated != null) {
+            DiscriminatorMetadata dmd = imd.getDiscriminatorMetadata();
+            if(dmd == null) dmd = imd.newDiscriminatorMetadata();
+            imd.setStrategy(InheritanceStrategy.NEW_TABLE);
+            dmd.setValue(definition.getName());
+            dmd.setStrategy(DiscriminatorStrategy.VALUE_MAP);
+            dmd.setColumn("DISCRIMINATOR");
+            LOGGER.info("fixDiscriminator: base: {}, {}", i, cmd.getName());
```
pgesek: these logs should be debug level

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
```diff
@@ -245,8 +289,7 @@ private void addInheritanceMetadata(ClassMetadata cmd, Class<?> definition) {
         Inheritance annotation = AnnotationUtils.findAnnotation(definition, ann);
 
         if (annotation == null) {
-            InheritanceMetadata imd = cmd.newInheritanceMetadata();
-            imd.setCustomStrategy("complete-table");
+            fixDiscriminator(definition, cmd, 2);
```
pgesek: ok, if we are calling this here, do we need this during generation?

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
```diff
@@ -144,6 +151,43 @@ public void addHelperClassMetadata(JDOMetadata jdoMetadata, ClassData classData,
         }
     }
 
+    private void fixDiscriminator(Class<?> definition, ClassMetadata cmd, int i) {
```
pgesek: What's the 'i' variable for?

platform/mds/mds/src/main/java/org/motechproject/mds/dto/EntityDto.java
```diff
@@ -40,9 +42,11 @@
     private Set<String> readOnlySecurityMembers;
     private String superClass;
     private boolean abstractClass;
+    private boolean isDiscriminated;
     private boolean securityOptionsModified;
     private Integer maxFetchDepth;
     private boolean readOnlyAccess;
+    private List<FieldDto> fieldsAdded;
```
pgesek: I would extend the name, so that it tells from where were the fields added (fromParent? fromChild?)

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
jefeweisen: Sorry, "this method mutates" is comment is obsolete.  Thx.

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -882,6 +883,40 @@ public String paramTypeForQuery() {
 
     @Override
     @Transactional
+    public List<EntityDto> findEntitiesBySuperclass(String classnameSuper) {
+        List<EntityDto> entities = new ArrayList<>();
+
+        FilterValue filterValue1 = new FilterValue() {
```
pgesek: Filters are generally only used for the filter part of the MDS data browser. There are better ways to do this, like simply using a query or one of the retrieve methods.

platform/mds/mds/src/main/java/org/motechproject/mds/repository/MotechDataRepository.java
```diff
@@ -138,6 +139,12 @@ public T retrieve(String[] properties, Object[] values, InstanceSecurityRestrict
         return (T) QueryExecutor.executeWithArray(query, values, restriction);
     }
 
+    public ForwardQueryResult retrieveSet(String[] properties, Object[] values, InstanceSecurityRestriction restriction) {
+        // avoids .setUnique
```
pgesek: I don't get this, who is doing .setUnique?

platform/mds/mds/src/main/java/org/motechproject/mds/repository/MotechDataRepository.java
```diff
@@ -177,9 +184,9 @@ public long count(InstanceSecurityRestriction restriction) {
     }
 
     public long count(String[] properties, Object[] values, InstanceSecurityRestriction restriction) {
-        Query query = createQuery(properties, values, restriction);
-        QueryUtil.setCountResult(query);
-        return (long) QueryExecutor.executeWithArray(query, values, restriction);
+        // This method is a workaround for avoiding .setUnique in .retrieve
```
pgesek: use retrieveAll then. Although, how is count doing setUnique? Isn't this a widely used method?

platform/mds/mds/src/main/java/org/motechproject/mds/service/EntityService.java
```diff
@@ -159,6 +161,8 @@
      */
     List<FieldDto> getFields(Long entityId);
 
+    List<FieldDto> getPossibleFields(Long entityId);
```
pgesek: What are possible fields?

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -924,7 +965,15 @@ public String paramTypeForQuery() {
...
 
+    private List<FieldDto> getFieldsPossible(Long entityId, boolean forDraft) {
+        return getFieldsI(entityId, forDraft, true);
+    }
+
     private List<FieldDto> getFields(Long entityId, boolean forDraft) {
+        return getFieldsI(entityId, forDraft, false);
+    }
+
+    private List<FieldDto> getFieldsI(Long entityId, boolean forDraft, boolean withPseudofields) {
```
pgesek: why the I?

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -964,9 +1014,68 @@ public int compare(Field o1, Field o2) {
...
+     * Adds display-only fields to the field list so that we can reuse UI widgets and
+     * wire protocol for editing entities with @Discriminated.
+     *
+     * One field is added for the discriminator.
+     * Once a discriminator is selected, the UI will expand the field list using fieldsAdded.
+     *
+     * Decode performed by:
+     *     org.motechproject.mds.web.service.impl.InstanceServiceImpl#DecodePseudofields
+     */
+    private void EncodePseudofields(Entity entity, List<FieldDto> fieldDtos) {
```
pgesek: follow java naming conventions

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -964,9 +1014,68 @@ public int compare(Field o1, Field o2) {
...
+        if(entity.isDiscriminated()) {
+            List<String> values = new ArrayList<String>();
+            List<EntityDto> entitiesDerived = new ArrayList<>();
+            int c=0;
```
pgesek: avoid variable like this

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -964,9 +1014,68 @@ public int compare(Field o1, Field o2) {
...
+            for (EntityDto e : entitiesDerived) {
+                List<FieldDto> fieldsAdded = new LinkedList<FieldDto>();
+                for (FieldDto fd2 : this.getFields(e.getId())) {
+                    if(!fieldids.contains(fd2.getBasic().getName()))
+                        fieldsAdded.add(fd2);
+                }
+                e.setFieldsAdded(fieldsAdded);
+            }
+
+            Type tD = allTypes.retrieveByClassName("java.util.Collection");
```
pgesek: the variable naming across this method should get improved

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -964,9 +1014,68 @@ public int compare(Field o1, Field o2) {
...
+            int c=0;
+            List<EntityDto> deriveds = this.findEntitiesBySuperclass(entity.getClassName());
+            if(deriveds != null) {
+                for (EntityDto derived : deriveds) {
+                    values.add(derived.getName());
+                    entitiesDerived.add(derived);
+                    c++;
+                }
+            }
+            Set<String> fieldids = new HashSet<String>();
```
pgesek: fieldIds

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -964,9 +1014,68 @@ public int compare(Field o1, Field o2) {
...
+     *
+     * Decode performed by:
+     *     org.motechproject.mds.web.service.impl.InstanceServiceImpl#DecodePseudofields
+     */
+    private void EncodePseudofields(Entity entity, List<FieldDto> fieldDtos) {
+        if(entity.isDiscriminated()) {
+            List<String> values = new ArrayList<String>();
+            List<EntityDto> entitiesDerived = new ArrayList<>();
+            int c=0;
+            List<EntityDto> deriveds = this.findEntitiesBySuperclass(entity.getClassName());
```
pgesek: drop 'this', we aren't using anywhere else

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
pgesek: and possibly other ways of clearing things up, I would see this logic getting split into smaller methods.

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/EntityServiceImpl.java
```diff
@@ -1200,6 +1309,11 @@ public LookupDto getLookupByName(Long entityId, String lookupName) {
     @Override
     @Transactional
     public List<FieldDto> getDisplayFields(Long entityId) {
+        return getDisplayFieldsI(entityId, true);
+    }
+
+    // TODO: inline
+    private List<FieldDto> getDisplayFieldsI(Long entityId, boolean andPossible) {
```
pgesek: andPossible?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/controller/EntityController.java
```diff
@@ -70,7 +70,7 @@
                     byModule.put(entity.getModule(), new ArrayList<EntityDto>());
                 }
 
-                if (!entity.isAbstractClass()) {
+                if (!entity.isAbstractClass() || entity.isDiscriminated()) {
```
pgesek: So why are we displaying these? Are we displaying all instances of entities inheriting this class?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jefeweisen: Extra fields crash the json mapper on the server side.  So for each field I either had to add the member, or delete it before it returns to the server.   This comments indicates the criterion for delete:

// {readonly, uiFilterable, . . .} = fields(FieldDto) - fields(FieldRecord)

What is the charter of FieldRecord?  Is there a good reason for FieldRecord to different from FieldDto?

pgesek: FieldRecord was geared towards the data browser, while fieldDto for the schema view. The only noteworthy difference I can see is that field records holds the value for the field. I think we could refactor FieldRecord to inherit from FieldDto in the future.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jefeweisen: Excellent question.  This is the reason for "TODO: remove hack".  It's not obvious.

The fieldsAdded are what I want to be in fieldsAdded: just the fields added based on knowing the discriminator.  Somehow fieldsAdded1 has been contaminated with the base Entity fields.  I need to make sure they are not present so the fields don't get duplicated.

I have a few suspicions of where it came from but after some work I have not isolated the true root cause.  I will take care of the root cause and remove the hack before merge.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jefeweisen: suffix T: "Temporary"
suffix I: "Internal"

Both are a symbol whose main significance is in avoiding a naming conflict.  Both should have a convention.  Above is a convention I have used before.  Please let me know if you have a preferred choice.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jefeweisen: Thanks for findCurrentScope().  I was unable to tell whether or not the existing code did a close enough action to work for my reuse.

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
jefeweisen: It should be called entityInstanceFields.

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
jefeweisen: generate* vs fix*: Thx.  Sounds good.

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
jefeweisen: Residual from testing.  Should be removed.  Thx.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jefeweisen: Thx.  Will change to for(var i=0;i<N;i=i+1) . . .

jefeweisen:
Any info on these would be helpful:

Does MOTECH have a well-defined supported browser list?  When I checked a couple of months ago I couldn't seem to find one.

Does MOTECH have a documented convention for class/method/variable naming?

pgesek:
@jefeweisen For naming see: http://docs.motechproject.org/en/latest/development/coding_conventions.htm
For the browser list see: http://docs.motechproject.org/en/latest/development/browser.html

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/controller/EntityController.java
jefeweisen: Re: "Why are we displaying these"

See https://applab.atlassian.net/browse/MOTECH-1649

    Widget appearance:
        - A non-editable label "subclass".
        - A mouse dropdown selector.
        - A choice for every y in Y, plus a choice for undefined.

    When creating an the, a widget allows the user to:

        1) See when y is has not been chosen.
        2) See the name of the subclass when y has been chosen.
        3) Choose y.

But the code fragment indicated is not any of this behavior.  The code fragment's behavior is:

    When the user updates an instance using entityId=x, we edit the instance the same as we would if we updated the instance with the entityId of its derived class.

Added to MOTECH-1649 for clarification.  Will also add a code comment.

platform/mds/mds/src/main/java/org/motechproject/mds/annotations/Discriminated.java
jefeweisen: Surprising: ". . .wondering about the purpose. . ."

Here's my best shot at clearing that up:  No I would not agree that the purpose is the UI behavior, though I could see how many (even most) might first be perceive the feature that way.

Also surprising: "Why not use JDO _ annotations?".

Maybe this helps? My answer to this is the same as my answer to:
  - Why @Entity?  Why not @PersistenceCapable?
  - Why have an Entity table?  Why not ask the database?
  - Why use MDS?  Why not use JDO?

Think of @Discriminated as the implementation of a datatype.  Originally there was a separate JIRA for it: https://applab.atlassian.net/browse/MOTECH-1768 though in practice it was not practical to develop separately.

Similarly, think of MOTECH-1649 as being Data Browser support for @Discriminated data.

jefeweisen: Incidentally, @Discriminated may be a case where naming matters.  I feel that "discriminated" may not be the best name on the grounds of false association with the JDO feature by the name "discriminator", coincidentally used for implementation.  Here are some alternatives I have considered:

@Alternative
@Union
@Choice

pgesek: I'm wondering this because we do have some general support for the regular JDO inheritance. In general we would use the regular JDO annotations unless we wanted our spin on it, for example with the entity annotation you get the databrowser, history etc. But on the other hand we just use the regular JDO unique annotation for annotating a column you want a unique constraint on. The main reasoning was that going with all new annotations would just be reinventing the wheel and needless work. I will admit though that we should consider moving some stuff to our annotations for clarity, "defaultFetchGroup" in @PersistenceCapable is an example. But back on point, I'm just wondering how we go forth with inheritance with MDS, will this be the recommended way, the only way or just an optional way to do inheritance?

Regarding the benefits, if you would ask my why use Entity instead of PersistenceCapable, I would say that you get the databrowser, history tracking, REST API and all that good stuff. Now, if I would ask use why use the new annotation instead of regular JDO inheritance, what would be the anwser? I primarily see UI benefits, namely the enhanced data browser support - a solution to that message campaing issue that was hotfixed for Ebodac using javascript customization. Is that correct or am I missing something?

As I mentioned, I see the value of this, I am just wondering how we go forward with inheritance support. Do we keep the two ways separate or do we merge them somehow?

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderImpl.java
pgesek: I've misread from where this was called, ignore the comment

platform/mds/mds/src/main/java/org/motechproject/mds/annotations/Discriminated.java
jefeweisen: Deciding what relationship MOTECH-1768 has to inheritance is a good question.  Here are a couple of points on MDS @Discriminator vs JDO @Inheritance:

1) My experience in metadata design is that verbatim inclusion of foreign metadata items (JDO @Inheritance) within a set of more tightly curated items (@Entity, @Field, all other MDS things) has a tendency to create pathologies about what combinations mean.  Often the pathologies are hard to identify up front, or only come into play when future changes are made.

2) Here is a dividing line I propose:

- Using @Discriminated means you can't also use @Inheritance on the same entity.
- If desired, sharing fields can be achived by java interface/implements, along with @Discriminated.  This presumes that the datanucleus enhancer can pass over the interfaces without getting confused.  Enhancer passing over seems plausible to me, but the datanucelus sources should probably be checked.

PR #118, opened 2015-10-12
Add evictOne method to MotechDataRepository

frankhuster:
In order to be more efficient when evicting one entity only


frankhuster:
Here you go, @pgesek

PR #119, opened 2015-10-12
MOTECH-1934: Fixed some issues with Scheduler actions

ngraczewski:
There were two issues with the task-channel.json from the Schedule module.

First one was that the "Schedule repeating job" action was passing the "repeatIntervalInMilliSeconds" argument as "LONG" type instead of the "INTEGER"

The second one was that the "Schedule repeating job with period interval" action was passing the "schedulePeriodRepeatingJob" as the "serviceMethod" parameter instead of the "scheduleRepeatingPeriodJob".

The last issue with the Scheduler action was that the repeating job with period interval was recognised as the repeating job and therefore there were problems with builders casting.


pgesek:
@ngraczewski Can you fix these two issues in the scheduler channel by the way? 

1) Event params map should not be required in scheduler actions, we should be able to create a job that only has a subject
2) End date should not be required, we can have neverending jobs.

PR #126, opened 2015-10-13
Added methods for evicting cache

pgesek:


frankhuster:
Much better than what I did in 0.27-X

PR #130, opened 2015-10-13
MOTECH-1937 Header and footer of the jqGrid always visible.

wstrzelczyk:
*From now, a user can see vertical scroll of the jqgrid even then display many records
 and cells have a large number of strings.
*Optimized CSS classes


pgesek:
We should distinguish the grid space more, perhaps gray out the unused space in a grid

PR #131, opened 2015-10-13
Catching up with Pawel's work on master branch

frankhuster:


platform/mds/mds/src/main/java/org/motechproject/mds/service/MotechDataService.java
```diff
@@ -218,14 +218,20 @@
...
+     * @param instance the instance to clear the cache for
+     */
+    void evictCacheForInstance(T instance);
+
+    /**
+     * Evicts cache for the entity class of this data service.
+     * @param withSubclasses if true, the cache for subclasses of the entity will be also cleared
      */
-    void evictOne(T object);
+    void evictEntityCache(boolean withSubclasses);
```
koshalt: not sure I like hiding the feature to evict sub classes inside an argument for the method call and would rather have it be 2 calls... but then again, it didn't kill anyone.

koshalt:
lgtm

PR #138, opened 2015-10-15
MOTECH-1942 SQL migration fixes 0.25 -> 0.26

sebbrudzinski:
Since version 0.26 the openId is marked as unique. This
means that empty strings will be recognized as duplicates
and won't be allowed. Until version 0.26 we used to put
empty string as openId if it wasn't provided. This migration
changes all empty openId fields to NULL since they are
allowed, even when unique constraint is placed. Moreover
some other migrations have been fixed.


kmadej:
Looks good to me, needs rebase

sebbrudzinski:
@pgesek tested and verified by Kamil

PR #142, opened 2015-10-16
Fix combobox tests on PostgreSQL

pgesek:
* Added ordering to retrieval of options.
  They will be returned alphabetically.


sebbrudzinski:
@pgesek OK to merge

PR #158, opened 2015-10-22
MOTECH-1972: Password Reminder - 500 Internal Server Error when log i…

tstalka:
…n with test user after changing password expiration settings

Now when the lastPasswordChange is null the modificationDate field will be used.


tstalka:
test this please

PR #161, opened 2015-10-22
MOTECH-1963 Disable sorting by relationship field on MDS Data Browser

smalecki:


pgesek:
The sorting works for serialized maps, (Map<?, Object>) i.e. in ModulePropertiesRecord. It won't work for maps that have separate tables, i.e. a Map<String,String> in the SettingsRecord. Moreover, it also doesn't work for comboboxes that have a separate table(multi-select comboboxes), i.e. permissions in MotechRole.

Would be best to disable sorting just for those maps don't work, but I can see just disabling it for all maps. For comboboxes, we should only disable it for multi select comboboxes.

pgesek:
#resolve

pgesek:
MOTECH-1963 resolve

pgesek:
MOTECH-1963 #resolve

PR #166, opened 2015-10-23
MOTECH-1869: Make creating an index for a lookup optional

shubhambeehyv:
Added indexRequired field in Lookup to make indexing optional.


shubhambeehyv:
test this please

platform/mds/mds/src/main/java/org/motechproject/mds/annotations/internal/LookupProcessor.java
```diff
@@ -113,6 +114,7 @@ protected void process(AnnotatedElement annotatedElement) {
         lookup.setLookupFields(lookupFields);
         lookup.setReadOnly(true);
         lookup.setMethodName(method.getName());
+        lookup.setIndexRequired(indexRequired);
```
pgesek: we should  extend the lookup processor test to check this

1stmateusz:
test this please

pgesek:
Moved to #327

PR #170, opened 2015-10-26
MOTECH-1940: Default UI tab for modules is determined based on permissions.

LukSkarDev:
- Added tabAccessMap to ModuleRegistrationData. This map contains permissions as keys and lists of available UI tabs for permission as values. Modules with more than one role can use it to register which role has access to which tabs.
- ModuleRegistrationData now determines default tab basing on delivered properties. If defaultURL is specified it will be used as URL of default tab (so all modules with only one role and specified defaultURL will still works). If not then tabAccessMap should be specified. ModuleRegistrationData will take first url tahat is accessible for current user permissions and use it as default tab URL.
- Added AvailabilityController (with test) to server-budnle. Modules that have used their own AvailabilityController now uses this one.


platform/server-bundle/src/main/java/org/motechproject/server/web/controller/AvailabilityController.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Controller class responsible for determining available UI tabs for modules where
+ * UI tabs availability depends on different user permissions.
+ */
+@Controller
+public class AvailabilityController {
```
sebbrudzinski: I believe this controller was supposed to determine available tabs, based on the data in ModuleRegistrationData, not hardcode settings for each module. What if external module wants to use it? They have to edit platform code?

platform/osgi-web-util/pom.xml
```diff
@@ -56,6 +56,10 @@
             <artifactId>org.motechproject.org.codehaus.jackson</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.springframework.security</groupId>
```
pgesek: Do we need this dependency? Why not have the code that determines the tab in server-bundle? Seems to belong more there

platform/osgi-web-util/src/main/java/org/motechproject/osgi/web/ModuleRegistrationData.java
```diff
@@ -207,6 +213,20 @@ public void setSubMenu(Map<String, SubmenuInfo> subMenu) {
...
+     */
+    @JsonIgnore
+    public Map<String, List<String>> getTabAccessMap() { return tabAccessMap; }
+
+    /**
+     * Sets the map available tabs for specified permissions for this module.
+     * @param tabAccessMap a map, where the keys are permissions names and values are lists of accessible tabs for that permission
+     */
+    @JsonIgnore
+    public void setTabAccessMap(Map<String, List<String>> tabAccessMap) { this.tabAccessMap = tabAccessMap; }
```
pgesek: I would rather expect tabs to be the keys, and permission list to be the values. Why do it in this backward fashion? Moreover, are we supporting a case where we have a tab without any permissions required? Flipping keys and values will probably be required for that.

platform/osgi-web-util/src/main/java/org/motechproject/osgi/web/ModuleRegistrationData.java
```diff
@@ -396,6 +416,30 @@ public String getDocumentationUrl() {
...
+    /**
+     * Determines which tab of module should be default and loaded as first based on
+     * user permissions.
+     *
+     * @return the name of the default tab
+     */
+    @JsonIgnore
+    public String determineDefaultTab() {
+
+        if(StringUtils.isNotBlank(defaultURL)) {
```
pgesek: Is this correct? don't we want to override this in case the user doesn't have the permission?

platform/server-bundle/src/main/java/org/motechproject/server/web/controller/AvailabilityController.java
```diff
@@ -0,0 +1,53 @@
...
+    @ResponseBody
+    public List<String> getAvailableTabs(@PathVariable String moduleName) {
+
+        Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+        Map<String, List<String>> tabAccessMap = uiFrameworkService.getModuleData(moduleName).getTabAccessMap();
+
+        ArrayList<String> availableTabs = new ArrayList<String>();
+        for (String tab : tabAccessMap.keySet()) {
+            //If no permissions were specified tab is available for everyone
+            if(tabAccessMap.get(tab).size() == 0) {
```
pgesek: we should treat null permissions the same as zero size

platform/server-bundle/src/main/java/org/motechproject/server/web/controller/AvailabilityController.java
```diff
@@ -0,0 +1,53 @@
...
+                availableTabs.add(tab);
+            }
+            for (String permission : tabAccessMap.get(tab)) {
```
pgesek: should be done in else, since I can see someone putting null in the map instead of an empty list

platform/server-bundle/src/main/java/org/motechproject/server/web/helper/MenuBuilder.java
```diff
@@ -184,4 +188,61 @@ private String getAngularModuleName(ModuleRegistrationData data) {
...
+                return buildDefaultLink(angularModule, tab);
+            }
+        }
+        return null;
+    }
+
+    private boolean isTabAvailable(String tab, Map<String, List<String>> tabAccessMap, Authentication auth) {
+
+        //If no permissions were specified tab is available for everyone
+        if (tabAccessMap.get(tab).size() == 0) {
```
pgesek: same as before, I would guard against null here

PR #181, opened 2015-11-02
MOTECH-1964: Validation for task name

LukSkarDev:
Added task name validation - the TaskNameAlreadyExistsException will be thrown when attempting to save task with already existing name. Added test for new behaviour.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/TaskNameAlreadyExistsException.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/TaskNameAlreadyExistsException.java
@@ -1,0 +1,10 @@
+package org.motechproject.tasks.ex;
+
+public class TaskNameAlreadyExistsException extends RuntimeException {
+
+    private static final long serialVersionUID = 9084172986189170522L;
+
+    public TaskNameAlreadyExistsException(String taskName) {
+        super("Task with name " + taskName + " already exists");
+    }
+}
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/TaskNameAlreadyExistsException.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/TaskNameAlreadyExistsException.java
@@ -1,9 +1,17 @@
 package org.motechproject.tasks.ex;
 
+/**
+ * Thrown when attempting to save task with already existing task name.
+ */
 public class TaskNameAlreadyExistsException extends RuntimeException {
 
     private static final long serialVersionUID = 9084172986189170522L;
 
+    /**
+     * Exception constructor.
+     *
+     * @param taskName the task name
+     */
     public TaskNameAlreadyExistsException(String taskName) {
         super("Task with name " + taskName + " already exists");
     }
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
@@ -584,14 +584,12 @@ public class TaskServiceImpl implements TaskService {
     private void validateName(Task task) {
 
         Long taskId = task.getId();
-        int tasksWithName = tasksDataService.findTasksByName(task.getName()).size();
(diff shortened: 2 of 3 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -579,6 +581,22 @@ private TaskDataProvider findProviderByName(DataSource ds) {
         return errors;
     }
 
+    private void validateName(Task task) {
+
+        Long taskId = task.getId();
+        int tasksWithName = tasksDataService.findTasksByName(task.getName()).size();
```
sebbrudzinski: Can we store the returned objects here, so we don't have to repeat database call in line 593?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -579,6 +581,22 @@ private TaskDataProvider findProviderByName(DataSource ds) {
...
+
+        if(taskId == null && tasksWithName > 0) {
+            LOGGER.warn("Task with name: {} already exists", task.getName());
+            throw new TaskNameAlreadyExistsException(task.getName());
```
sebbrudzinski: Log and throw is an anti-pattern; either log or throw.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -579,6 +581,22 @@ private TaskDataProvider findProviderByName(DataSource ds) {
...
+        } else if (taskId != null && tasksWithName > 0) {
+            if(!(tasksDataService.findTasksByName(task.getName()).get(0).getId().equals(taskId))) {
+                LOGGER.warn("Task with name: {} already exists", task.getName());
+                throw new TaskNameAlreadyExistsException(task.getName());
```
sebbrudzinski: Log and throw is an anti-pattern; either log or throw.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/TaskNameAlreadyExistsException.java
```diff
@@ -0,0 +1,10 @@
+package org.motechproject.tasks.ex;
+
+public class TaskNameAlreadyExistsException extends RuntimeException {
```
sebbrudzinski: Needs javadoc

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskServiceImplTest.java
```diff
@@ -192,6 +193,16 @@ public void shouldNotSaveTaskIfNotMatchExtraValidationConditions() {
...
     }
 
+    @Test(expected = TaskNameAlreadyExistsException.class)
+    public void shouldNotSaveTaskWithDuplicateName() {
+
+        when(tasksDataService.findTasksByName("name")).thenReturn(
+                asList(new Task("name", trigger, asList(action))));
+
+        Task t = new Task("name", trigger, asList(action), null, false, false);
+        taskService.save(t);
```
sebbrudzinski: Can we also verify in the finally block, that there were no more interactions with the TasksDataService?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -579,6 +581,20 @@ private TaskDataProvider findProviderByName(DataSource ds) {
...
 
+    private void validateName(Task task) {
+
+        Long taskId = task.getId();
+        List<Task> tasksWithName = tasksDataService.findTasksByName(task.getName());
+
+        if(taskId == null && tasksWithName.size() > 0) {
+            throw new TaskNameAlreadyExistsException(task.getName());
+        } else if (taskId != null && tasksWithName.size() > 0) {
+            if(!(tasksWithName.get(0).getId().equals(taskId))) {
```
sebbrudzinski: Just move it to the if the line above and get rid of unnecessary brackets

PR #183, opened 2015-11-04
MOTECH-1954: Added possibility to delete drag and drop bubbles 

tstalka:
- each bubble can be deleted by the "X" button
- fixed the magical disappearance of manipulation
- fixed manipulation dropdown


modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -695,7 +718,7 @@
                         trigger: 'manual'
                     }).click(function (event) {
                         event.stopPropagation();
-                        if (!$(this).hasClass('hasPopoverShow')) {
+                        if (!$(this).hasClass('hasPopoverShow') && (event.target === undefined || event.target.className !== 'close badge-close')) {
```
pgesek: why not just do event.target || ...

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -705,18 +728,27 @@
                                 otherPopoverElem.removeClass('hasPopoverShow');
                                 otherPopoverElem.removeAttr('ismanipulate');
                             }
-                            if (filter.key) {
+                            if (filter !== undefined && filter.key) {
```
pgesek: why not just do filter && ...

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -705,18 +728,27 @@
...
                                 $(this).attr('manipulate', filter.key.split("?").slice(1).join(" "));
                             }
 
                             $(this).addClass('hasPopoverShow');
                             $(this).attr('ismanipulate', 'true');
                             $(this).popover('show');
-                        } else {
+                        } else if (event.target === undefined || event.target.className !== 'close badge-close') {
```
pgesek: same as above

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -737,7 +769,7 @@
                     });
 
                     el.on("manipulateChanged", function () {
-                        if (filter.key) {
+                        if (filter !== undefined && filter.key) {
```
pgesek: same as above

tstalka:
test this please

PR #184, opened 2015-11-05
MOTECH-1269: Refactor history and trash handling

pgesek:
* made history/trash use InstanceLifecycleListeners
* history and trash now store relations as collections of numbers (ids), they no longer copy the whole tree
* made history/trash just copy properties without making queries for the db schema
* added DDE ITs for history handling


sebbrudzinski:
@pgesek OK to merge

PR #185, opened 2015-11-05
MOTECH-1970 Added backend support for inline relationships view

sebbrudzinski:
As a part of this commit backend support for viewing
relationships has been implemented. New method has
been added in the MotechDataService, that allows to
fetch a collection of instances with passed IDs with
one call.


platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -692,61 +714,68 @@ private void setRelationProperty(Object instance, FieldRecord fieldRecord) throw
...
+           Collection relatedInstances = ((Collection) parsedValue);
 
-           for (Object object : (Collection) fieldValue) {
-               if (isFromUI(object)) {
-                   ((Collection) parsedValue).add(findRelatedObjectById(((Map) object).get(ID_FIELD_NAME), service));
-               } else if (isHistoricalObject(object)) {
-                   String currentVersion = HistoryTrashClassHelper.currentVersion(object.getClass());
-                   ((Collection) parsedValue).add(findRelatedObjectById(PropertyUtil.safeGetProperty(object, currentVersion), service));
+           // Add already related instances, provided any exist
+           if (related != null) {
```
pgesek: 1) Change the naming, what's the difference between related and relatedInstances?
2) If we assume related is a collection, do that at the method signature level

platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java
```diff
@@ -414,6 +415,15 @@ public T findById(Long id) {
     }
 
     @Override
+    public Collection<T> findByIds(Collection<Long> ids) {
+        if (ids == null) {
+            return null;
```
pgesek: better return an empty list or throw illegalArgumentException, also any reason for this returning collection since our other methods return lists?

sebbrudzinski:
Merged into branch on upstream repo. Branch: dev/inline-relationships

PR #188, opened 2015-11-06
MOTECH-1979: Fix for regenerating the mds-entities.jar

tstalka:
- Removed tenant id
- MDS class loader will be used as the context class loader


platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/JarGeneratorServiceImpl.java
```diff
@@ -186,6 +189,7 @@ private synchronized void regenerateMdsDataBundle(boolean startBundle, String...
 
         // Give framework some time before returning to the caller
         ThreadSuspender.sleep(2000);
+        Thread.currentThread().setContextClassLoader(currentClassLoader);
```
pgesek: Do this in finally

PR #189, opened 2015-11-09
MOTECH-1974 Include object specific parameters as a part of the MotechEvent class

martokarski:


platform/event/src/main/java/org/motechproject/event/MotechEvent.java
```diff
@@ -70,6 +73,60 @@ public void setId(UUID id) {
...
+     * Sets event as discarded
+     *
+     * @param value
+     */
+    public void setDiscarded(boolean value) {
+        discarded = value;
+    }
+
+    /**
+     * Returns whether event is broadcast
```
pgesek: a broadcast event

platform/event/src/main/java/org/motechproject/event/MotechEvent.java
```diff
@@ -70,6 +73,60 @@ public void setId(UUID id) {
...
+     *
+     * @return broadcast
+     */
+    public boolean isBroadcast() {
+        return broadcast;
+    }
+
+    /**
+     * Sets event as broadcast
```
pgesek: a broadcast event

platform/event/src/main/java/org/motechproject/event/MotechEvent.java
```diff
@@ -79,6 +136,24 @@ public String getSubject() {
     }
 
     /**
+     * Returns destination of the message
```
pgesek: - the id of the Motech listener that this event is meant for

platform/event/src/main/java/org/motechproject/event/MotechEvent.java
```diff
@@ -79,6 +136,24 @@ public String getSubject() {
...
+     *
+     * @return message destination
+     */
+    public String getMessageDestination() {
+        return messageDestination;
+    }
+
+    /**
+     * Sets message destination
```
pgesek: - the id of the Motech listener that this event is meant for

platform/event/src/main/java/org/motechproject/event/listener/proxy/EventAnnotationBeanPostProcessor.java
```diff
@@ -142,7 +142,7 @@ public void clearListeners(ApplicationContext applicationContext) {
         }
     }
 
-    //TODO:keeping required false for now, should be removed.
+    // @TODO:keeping required false for now, should be removed.
     @Autowired(required = false)
```
pgesek: why the @? btw, I think we should be good to remove required = false now anyway

pgesek:
You will have to also create a commit for modules that removes the usage of the removed constants from the DHIS module.

platform/event/src/main/java/org/motechproject/event/MotechEvent.java
```diff
@@ -15,14 +16,16 @@
  */
 public class MotechEvent implements Serializable {
     public static final String EVENT_TYPE_KEY_NAME = "eventType";
```
pgesek: What's event type? Is this for the scheduler? If its only for the scheduler, we should move this to scheduler module, the scheduler service interface for example.

sebbrudzinski:
LGTM

PR #190, opened 2015-11-09
MOTECH-1961 Added logging in MotechAccessVoter

sebbrudzinski:


platform/web-security/src/main/java/org/motechproject/security/authentication/MotechAccessVoter.java
```diff
@@ -45,8 +50,11 @@ public boolean supports(Class<?> clazz) {
...
         int result = ACCESS_ABSTAIN;
 
+        LOGGER.debug("Voting on the access to the {} for user {} with permissions {}", object, authentication.getPrincipal(), authentication.getAuthorities());
+
         for (ConfigAttribute attribute : attributes) {
             if (this.supports(attribute)) {
+                LOGGER.debug("Verifying supported attribute: " + attribute.getAttribute());
                 result = ACCESS_DENIED;
 
                 if (StringUtils.equalsIgnoreCase(SecurityConfigConstants.USER_ACCESS_PREFIX + authentication.getName(),
```
pgesek: maybe log something if we land on the return in this if

PR #192, opened 2015-11-10
MOTECH-1976: Active style for tabs is applied correctly

LukSkarDev:
Added selectedTabState to server-bundle, which allows to correctly apply 'active' style for tabs with redirect. Changed the way in which 'active' style is set for tabs and breadcrumb to work with new approach. Removed active() method from server-bundle as it is not needed anymore.


sebbrudzinski:
http-bundle archetype should be updated with the latest changes as well. Also, I'm assuming all custom built modules will have to adjust their UI as well, if they used the old active function?

sebbrudzinski:
Something is still not quite right with the http archetype. The default tab is not active when I enter the built module.

PR #193, opened 2015-11-10
MOTECH-1978: cast webapploader to its superclass

1stmateusz:
in tomcat-7 WebappClassLoader has field "resourceEntries" but in tomcat-8
only its superclass has this field so the object must be casted to its
superclass to get resourceEntries field


motech-gerrit:

motech-gerrit:

motech-gerrit:

PR #194, opened 2015-11-10
MOTECH-1978: cast WebappClassLoader to its superclass

1stmateusz:
in tomcat-7 WebappClassLoader has field "resourceEntries" but in tomcat-8 only its superclass has this field so the object must be casted to its superclass to get "resourceEntries" field

### DIFF
--- a/platform/server/src/main/java/org/motechproject/server/impl/JspBundleLoader.java
+++ b/platform/server/src/main/java/org/motechproject/server/impl/JspBundleLoader.java
@@ -2,6 +2,7 @@ package org.motechproject.server.impl;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.reflect.FieldUtils;
 import org.motechproject.server.api.BundleLoader;
 import org.motechproject.server.api.BundleLoadingException;
 import org.osgi.framework.Bundle;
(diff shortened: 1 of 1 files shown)


motech-gerrit:

motech-gerrit:

pgesek:
test this please

platform/server/src/main/java/org/motechproject/server/impl/JspBundleLoader.java
```diff
@@ -204,7 +206,19 @@ private static void clearTomcatCache() throws InvocationTargetException, NoSuchM
         Class cl = loader.getClass();
 
         if ("org.apache.catalina.loader.WebappClassLoader".equals(cl.getName())) {
-            clearMap(cl, loader, "resourceEntries");
+            ArrayList<Field> fields = new ArrayList<Field>(Arrays.asList(cl.getDeclaredFields()));
```
pgesek: why do we need the array list? and the list from arrays? why not just iterate over the array?

Anyway, probably would be best to just use FieldUtils.getDeclaredField() from apache commons-lang.

1stmateusz:
FieldUtils.getDeclaredField() doesn't return a required field (don't know why), so I stayed with casting object to superclass.

sebbrudzinski:
We can use ```static List<Field> getAllFieldsList(Class<?> cls)``` to get a list of all fields, even those from parent classes

1stmateusz:
I used static Field getField(Class cls, String fieldName) to get field from object superclasses and it works fine.

sebbrudzinski:
test this please

PR #195, opened 2015-11-13
MOTECH-1969: Clicking the drop down button should close the lookup panel

atish160384:
* Removed an 'if' condition in the mouseup function for the 'document'

### DIFF
--- a/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
@@ -4632,9 +4632,7 @@
 
             if (!container.is(e.target) && container.has(e.target).length === 0 &&
                 !button.is(e.target) && button.has(e.target).length === 0 && container.is(":visible")) {
-                if (e.target.offsetParent.hasAttribute("id") && e.target.offsetParent.id !== "ui-datepicker-div") {
-                    $scope.showLookupDialog();
-                }
+                $scope.showLookupDialog();
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4632,9 +4632,7 @@
 
             if (!container.is(e.target) && container.has(e.target).length === 0 &&
                 !button.is(e.target) && button.has(e.target).length === 0 && container.is(":visible")) {
-                if (e.target.offsetParent.hasAttribute("id") && e.target.offsetParent.id !== "ui-datepicker-div") {
-                    $scope.showLookupDialog();
-                }
```
wstrzelczyk: You can not remove this 'if' condition, because this condition prevents closing lookups panel when user operate on datepicker and also when user wants to close the datepicker panel by clicking outside of this panel.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4628,11 +4628,11 @@
         */
         $(document).mouseup(function (e) {
             var container = $("#lookup-dialog"),
-            button = $("#lookupDialogButton");
+            button = $("#lookupDialogButton"),date_picker = $("#ui-datepicker-div");
```
wstrzelczyk: should be divided into two line

pgesek:
test this please

wstrzelczyk:
ok, ready to merge

PR #196, opened 2015-11-16
MOTECH-1861: Refactored MDS to use dtos when processing schema

pgesek:
I have switched MDS to use dtos instead of domain objects when building classes and
metadata. This limits database queries and minimalizes what has to be done inside a transaction,
which is a good thing. I had to change how textareas are handled a bit, since some of their UI specific
processing was woven into the dto classes, which we now use when processing.

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
@@ -16,9 +16,9 @@ import org.motechproject.mds.domain.ClassData;
 import org.motechproject.mds.domain.ComboboxHolder;
 import org.motechproject.mds.domain.Entity;
 import org.motechproject.mds.domain.EntityType;
-import org.motechproject.mds.dto.SchemaHolder;
 import org.motechproject.mds.dto.EntityDto;
 import org.motechproject.mds.dto.FieldDto;
+import org.motechproject.mds.dto.SchemaHolder;
 import org.motechproject.mds.dto.TypeDto;
 import org.motechproject.mds.enhancer.MdsJDOEnhancer;
 import org.motechproject.mds.ex.entity.EntityCreationException;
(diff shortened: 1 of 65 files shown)


platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
```diff
@@ -65,8 +67,10 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MDSConstructorImpl.class);
 
-    private MdsConfig mdsConfig;
+    @Autowired
     private AllEntities allEntities;
```
sebbrudzinski: Why do we initialize one bean on the field level and another on the setter level?

platform/mds/mds/src/main/java/org/motechproject/mds/dto/TypeValidationDto.java
```diff
@@ -0,0 +1,52 @@
+package org.motechproject.mds.dto;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+
```
sebbrudzinski: We will need javadoc here as well

platform/mds/mds/src/main/java/org/motechproject/mds/dto/UIDisplayFieldComparator.java
```diff
@@ -0,0 +1,50 @@
+
```
sebbrudzinski: Fix empty line

platform/mds/mds/src/main/java/org/motechproject/mds/helper/JavassistBuilder.java
```diff
@@ -109,18 +109,29 @@ public static CtMethod createSetter(String fieldName, CtField field) throws Cann
...
         } else {
             return createInitializerForThirdPartyPackages(typeClass, defaultValue);
         }
     }
 
+
     private static CtField.Initializer createInitializerForJavaPlatformPackages(String typeClass,
                                                                                 Object defaultValue,
-                                                                                String defaultValueAsString) {
+                                                                                Object defaultValueg) {
```
sebbrudzinski: Valueg? Should probably be renamed

platform/mds/mds/src/main/java/org/motechproject/mds/service/EntityService.java
```diff
@@ -485,4 +486,6 @@
      * @return the list of fields for the UI
      */
     List<FieldDto> getEntityFieldsForUI(Long entityId);
+
+    SchemaHolder getSchema();
```
sebbrudzinski: Needs javadoc

sebbrudzinski:
@pgesek LGTM

PR #197, opened 2015-11-16
 MOTECH-1800: UserPreferences for MDS entities

tstalka:
- Added global settings for grid size
- Added possibility to store data about visible fields in the database
- User can have other preferences for each entity
- Cherry-pick: Refactor of the EntityControllerTest


platform/mds/mds-web/src/main/java/org/motechproject/mds/web/domain/UserPreferencesFields.java
```diff
@@ -0,0 +1,35 @@
+package org.motechproject.mds.web.domain;
+
+/**
+ * The <code>UserPreferencesFields</code> contains data about visible fields update.
```
pgesek: a visible fields update

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/domain/UserPreferencesFields.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class UserPreferencesFields {
+
+    private String field;
+
+    private String action;
```
pgesek: why not use an enum here?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/domain/UserPreferencesFields.java
```diff
@@ -0,0 +1,35 @@
+package org.motechproject.mds.web.domain;
+
+/**
+ * The <code>UserPreferencesFields</code> contains data about visible fields update.
+ */
+public class UserPreferencesFields {
```
pgesek: Rename to GridFieldSelectionUpdate

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -143,18 +143,33 @@
...
+                if (name === selectedFields[i].basic.name) {
+                    return true;
                 }
-                return (!result);
+            }
+        }
+        return false;
+    }
+
+    function handleGridPagination(pgButton, pager, scope, http) {
```
pgesek: Don't like the look of this function to be honest, it could use some comments, so that I have at least some clue what's happening here. Moreover, why not just save the page size in the retrieve records function of the controller? That way we can avoid an extra http call, and if we just do an update if the setting changed then this shouldn't affect performance much.

platform/mds/mds/src/main/java/org/motechproject/mds/config/ModuleSettings.java
```diff
@@ -159,6 +162,30 @@ public void setTimeUnit(TimeUnit timeUnit) {
...
+
+    public void setGridSize(Integer size) {
+        set(MDS_DEFAULT_GRID_SIZE, size, DEFAULT_GRID_SIZE);
+    }
+
+    public Integer getGridSize() {
+        Object objectValue = getValueForKey(MDS_DEFAULT_GRID_SIZE);
+        Integer value = DEFAULT_GRID_SIZE;
+
+        if (objectValue != null && StringUtils.isNotBlank(objectValue.toString())) {
```
pgesek: do we need this? the setters already parse to integer

platform/mds/mds/src/main/java/org/motechproject/mds/config/ModuleSettings.java
```diff
@@ -215,6 +242,7 @@ public void serialize(ModuleSettings value, JsonGenerator jgen, SerializerProvid
             jgen.writeObjectField("emptyTrash", value.isEmptyTrash());
             jgen.writeObjectField("timeValue", value.getTimeValue());
             jgen.writeObjectField("timeUnit", value.getTimeUnit());
+            jgen.writeObjectField("gridSize", value.getGridSize());
```
pgesek: off topic, but what's the reason for all this? why isn't this class just a pojo, but stores everything in a map instead?

platform/mds/mds/src/main/java/org/motechproject/mds/service/UserPreferencesService.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public interface UserPreferencesService {
+
+    /**
+     * Returns preferences for entity with the given id.
+     *
+     * @param id the id of  the entity for the preferences
+     * @return list of user preferences for the entity
+     */
+    List<UserPreferencesDto> getEntityPreferences(Long id);
```
pgesek: rename the id parameter to entityId, everywhere where it applies

platform/mds/mds/src/main/java/org/motechproject/mds/service/UserPreferencesService.java
```diff
@@ -0,0 +1,84 @@
...
+     * created.
+     *
+     * @param id the id of  the entity for the preferences
+     * @param username the owner of the preferences
+     * @param fieldName the name of the field to add
+     */
+    void selectField(Long id, String username, String fieldName);
+
+    /**
+     * Removes field with the given name from the user visible fields. If preferences doesn't exist then default will be
```
pgesek: don't

pgesek: default ones

pgesek: those two comments apply to all javadocs here

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/UserPreferencesServiceImpl.java
```diff
@@ -0,0 +1,199 @@
...
+    public void selectField(Long id, String username, String fieldName) {
+        Entity entity = getEntity(id);
+        UserPreferences userPreferences =  allUserPreferences.retrieveByClassNameAndUsername(entity.getClassName(), username);
+        userPreferences = checkPreferences(userPreferences, entity, username);
+
+        Field field = entity.getField(fieldName);
+        assertField(field);
+        List<Field> fields = userPreferences.getVisibleFields();
+        fields.add(field);
+        userPreferences.setVisibleFields(fields);
```
pgesek: set is probably not required since you are manipulating the orignal collection (maybe it would be better to just add an addField method to UserPreferences though)

platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/UserPreferencesServiceImpl.java
```diff
@@ -0,0 +1,199 @@
...
+        Entity entity = getEntity(id);
+        UserPreferences userPreferences =  allUserPreferences.retrieveByClassNameAndUsername(entity.getClassName(), username);
+        userPreferences = checkPreferences(userPreferences, entity, username);
+
+        Field field = entity.getField(fieldName);
+        assertField(field);
+        if (userPreferences.getVisibleFields().contains(field)) {
+            List<Field> fields = userPreferences.getVisibleFields();
+            fields.remove(field);
+            userPreferences.setVisibleFields(fields);
```
pgesek: same as above

platform/mds/mds/src/main/resources/db/migration/default/V40_1__MOTECH-1800.sql
```diff
@@ -0,0 +1,39 @@
...
+CREATE TABLE IF NOT EXISTS "ConfigSettings" (
+  "id" bigint,
+  "afterTimeUnit" varchar(255) NOT NULL,
+  "afterTimeValue" int,
+  "deleteMode" varchar(255) NOT NULL,
+  "emptyTrash" bit(1),
+  PRIMARY KEY ("id")
+);
+
+DO
```
pgesek: any reason for this? we should be safe to just add the field

platform/mds/mds/src/main/resources/db/migration/mysql/V40_1__MOTECH-1800.sql
```diff
@@ -0,0 +1,45 @@
...
+CREATE TABLE IF NOT EXISTS ConfigSettings (
+  id bigint(20),
+  afterTimeUnit varchar(255) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL,
+  afterTimeValue int(11),
+  deleteMode varchar(255) CHARACTER SET latin1 COLLATE latin1_bin NOT NULL,
+  emptyTrash bit(1),
+  PRIMARY KEY (id)
+);
+
+DELIMITER $$
```
pgesek: same as above, how can this field exist?

pgesek: also do we need the _1, if we are running migrations out of order now? Can't we just the number from master?

platform/mds/mds/src/main/resources/motech-mds.properties
```diff
@@ -3,3 +3,5 @@ mds.deleteMode = trash
 mds.emptyTrash = false
 mds.emptyTrash.afterTimeValue = 1
 mds.emptyTrash.afterTimeUnit = Hours
+
+mds.default.gridSize = 10
```
pgesek: Change this to 50

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -191,6 +210,111 @@ public void testEntitiesBundleInstallsProperly() throws Exception {
...
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someInteger"));
+
+        userPreferencesService.selectField(entityDto.getId(), USERNAME, "someString");
+        userPreferencesDto = userPreferencesService.getUserPreferences(entityDto.getId(), USERNAME);
+        assertEquals(9,  userPreferencesDto.getVisibleFields().size());
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someBoolean"));
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someString"));
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someInteger"));
+
+        // if field will be removed from entity the it should be removed also from preferences (CASCADE)
```
pgesek: , then

platform/mds/mds/src/test/java/org/motechproject/mds/it/osgi/MdsBundleIT.java
```diff
@@ -191,6 +210,111 @@ public void testEntitiesBundleInstallsProperly() throws Exception {
...
+        });
+
+        userPreferencesDto = userPreferencesService.getUserPreferences(entityDto.getId(), USERNAME);
+        assertEquals(8, userPreferencesDto.getVisibleFields().size());
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someBoolean"));
+        assertTrue(userPreferencesDto.getVisibleFields().contains("someString"));
+        assertFalse(userPreferencesDto.getVisibleFields().contains("someInteger"));
+    }
+
+    private EntityDto createEntity() {
```
pgesek: rename this to createEntityForPreferencesTest

platform/mds/mds/src/main/resources/db/migration/default/V40_1__MOTECH-1800.sql
tstalka: Migration has not been added in the past, so tabel is created when first record is created, in this case when we save settings in the Mds Settings tab.

pgesek:
There is one problem with the solution for storing visible fields - it always stores all the fields.

The problematic scenario is this:
1) Go to an Entity
2) Don't do anything with regards to the fields being displayed
3) Change the displayable fields in the entity schema
4) Go back to the databrowser - the change is not affecting the fields being displayed, since we have a list of visible fields stored from the first visit

My solution would be doing something similar to what we did with cookies if I remember correctly - only store the fields the user selected or unselected, then merge that with the default diplayable fields of the entity by adding fields selected manually and removing fields unselected manually.

PR #199, opened 2015-11-17
MOTECH-1982 Added event signaling that all modules have started

martokarski:


platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -20,13 +25,34 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PlatformStatusManagerImpl.class);
 
+    private static final String EVENT_SUBJECT = "All bundles started.";
```
pgesek: 1) this should be a public constant in PlatformConstants
2) the subject should be named according to convention - org/motechproject/osgi/event/XXX , XXX can be MODULES_STARTED for example

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -20,13 +25,34 @@
...
+    private static final String ERROR_MESSAGE = "OsgiEventProxy is set null";
```
pgesek: 1) Rename the variable so that it indicates what error this talks about - since the message is specific to OsgiEventProxy being null
2) set -> set to

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -20,13 +25,34 @@
...
+    private List<Bundle> blueprintBundles;
+
+    private OsgiEventProxy osgiEventProxy;
+
+    public PlatformStatusManagerImpl(List<Bundle> osgiBundles, List<Bundle> blueprintBundles) {
+        this.osgiBundles = osgiBundles;
+        this.blueprintBundles = blueprintBundles;
+    }
+
+    public PlatformStatusManagerImpl() {
```
pgesek: do we need this constructor?

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -87,7 +118,33 @@ private void handleContextFailedEvent(OsgiBundleContextFailedEvent event) {
...
+    }
+
+    private void checkBlueprintBundle(Bundle bundle) {
+        if(blueprintBundles.contains(bundle)) {
+            blueprintBundles.remove(bundle);
+            checkStartedBundle();
+        }
+    }
+
+    private void checkStartedBundle() {
```
pgesek: missleading method name, this doesn't check if a bundle is started, but checks if all bundles have started

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -87,7 +118,33 @@ private void handleContextFailedEvent(OsgiBundleContextFailedEvent event) {
         LOGGER.debug("Received context closed event {} from {}", event, symbolicName);
         LOGGER.info("{} failed to start due to {}", symbolicName, failureCauseMsg);
 
+        checkBlueprintBundle(event.getBundle());
+
         platformStatus.removeStartedBundle(symbolicName);
         platformStatus.addContextError(symbolicName, failureCauseMsg);
     }
+
+    private void checkOsgiBundle(Bundle bundle) {
```
pgesek: I would rename this to checkBundleOsgiStarted

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -87,7 +118,33 @@ private void handleContextFailedEvent(OsgiBundleContextFailedEvent event) {
...
+        if (osgiBundles.contains(bundle)) {
+            osgiBundles.remove(bundle);
+            checkStartedBundle();
+        }
+    }
+
+    private void checkBlueprintBundle(Bundle bundle) {
```
pgesek: would rename this to checkBundleBlueprintStarted

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -34,6 +60,7 @@ public void bundleChanged(BundleEvent bundleEvent) {
         if (bundleEvent.getType() == BundleEvent.STARTED) {
             LOGGER.debug("Bundle {} started", symbolicName);
             platformStatus.addOSGiStartedBundle(symbolicName);
+            checkOsgiBundle(bundleEvent.getBundle());
         } else if (bundleEvent.getType() == BundleEvent.STOPPED) {
             LOGGER.trace("Bundle {} stopped", symbolicName);
```
pgesek: shouldn't we handle this event as well? Are we handling errors?

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -5,11 +5,16 @@
...
+import org.motechproject.server.osgi.event.OsgiEventProxy;
+import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleEvent;
 import org.osgi.framework.BundleListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.List;
+
```
pgesek: The unit test for this class should get extended

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -34,14 +55,22 @@ public void bundleChanged(BundleEvent bundleEvent) {
         if (bundleEvent.getType() == BundleEvent.STARTED) {
             LOGGER.debug("Bundle {} started", symbolicName);
             platformStatus.addOSGiStartedBundle(symbolicName);
+            checkBundleOsgiStarted(bundleEvent.getBundle());
         } else if (bundleEvent.getType() == BundleEvent.STOPPED) {
             LOGGER.trace("Bundle {} stopped", symbolicName);
             platformStatus.removeOSGiStartedBundle(symbolicName);
+            osgiBundles.remove(bundleEvent.getBundle());
```
pgesek: why are we just removing and not calling the check method here? If this happens for the last bundle in line, the event will not be fired.

platform/osgi-platform/src/main/java/org/motechproject/server/osgi/status/PlatformStatusManagerImpl.java
```diff
@@ -34,14 +55,22 @@ public void bundleChanged(BundleEvent bundleEvent) {
...
+            blueprintBundles.remove(bundleEvent.getBundle());
+        } else if (bundleEvent.getType() == BundleEvent.UNRESOLVED) {
```
pgesek: Lets skip processing of unresolved for now, I have debug this, and the issue is that MDS refreshing the bundles to use new domain classes will produce this state, which is not an error indicator. I think for now we can just check the stopped osgi event and error blueprint events

PR #213, opened 2015-11-21
 MOTECH-1861: Reworking startup of entities context to limit transactions 

pgesek:
This should speed up MOTECH startup. Instead of going to the db for each service, we now read the schema from inside the bundle. Entity info is schema is included as json files inside the mds-entities
bundle.


sebbrudzinski:
@pgesek LGTM

PR #215, opened 2015-11-23
MOTECH-1628 MDS Data Browser UI: Inline relationships

sebbrudzinski:
This pull request adds the ability to add, edit and remove relationships of an
instance, directly from a parent entity. The UI has been improved to present
related instances in a grid, or as embeded fields, if the relationship is 1:1.
Some backend code has also been adjusted and refactored, in order to 
work with the new views.


sebbrudzinski:
test this please

sebbrudzinski:
test this please

sebbrudzinski:
test this please

sebbrudzinski:
1. We should not display negative (tempororary IDs in the relationship grids)
2. When I add a new relationship record using "Add new (entity name)", after I choose to add that related intance, the modal should dissapear. Currently it clears everything and stays opened with title only.
3. When I have more than one relationship in my entity, all the "Add new (entity name)" buttons open modal for one related entity only. For example, I've got relationshipField1, relationshipField2, relationshipField3 - I choose to add new relationship to relationshipField3, but the opened modal comes from relationshipField1

wstrzelczyk:
I fixed the problems reported above

pgesek:
Needs merge with master

pgesek:
I tried using this with the Task entity on Firefox and it is falling apart:

1) I can add trigger, but I still see an empty modal after I do
2) I can't add anything else, I am shown an empty add trigger modal no matter what I add 
3) If I try to add a Task Config the grids get stuck on loading

pgesek:
Nevermind, I had an old version

pgesek:
Can we do something with the loading? When I click add new task, it looks bad when the grids are loading

nickdotreid:
Are you talking about an MDS tasks instance or the Tasks UI?

In the tasks UI I'm planning to change the routes to use route.resolve...
so that slow data loads happen *before* the UI changes... then the data
gets directed directly into the controller

See: https://docs.angularjs.org/api/ngRoute/provider/$routeProvider

I'm not sure where I'll throw the UI-blocking code... but I bet there is a
nice event somewhere...

pgesek:
@nickdotreid I'm talking about adding a new task in MDS. When I do that, the grids for related field collection start loading sort of on top of the greyed out area, which does not look good imo.

pgesek:
The issue is fixed, however there exists with clearing field types.

Repro: 

1) Go to 'Add task' in MDS
2) Without refreshing go to 'Add MotechUser' in MDS

The MotechUser fields such as external id are now treated as relationship fields, which is not correct. Doing anything with the fields fails.

pgesek:
Remove the comments

PR #216, opened 2015-11-23
MOTECH-1961 Added logging in MotechDelegatingFilterProxy to track requests

sebbrudzinski:


platform/web-security/src/main/java/org/motechproject/security/filter/MotechDelegatingFilterProxy.java
```diff
@@ -60,6 +66,10 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         WebApplicationContext context = super.findWebApplicationContext();
         MotechProxyManager proxyManager = context.getBean(MotechProxyManager.class);
 
+        if (request instanceof HttpServletRequest) {
+            traceRequest((HttpServletRequest) request);
+        }
```
pgesek: is an else possible? Maybe log someting if its a different type of request

platform/web-security/src/main/java/org/motechproject/security/filter/MotechDelegatingFilterProxy.java
```diff
@@ -97,6 +107,15 @@ private boolean checkAdminMode() throws IOException {
         return adminModeProperty;
     }
 
+    private void traceRequest(HttpServletRequest req) {
+        LOGGER.trace("HTTP request {} received from {}", req.getPathInfo(), req.getRemoteAddr());
+        if (req.getSession() != null && req.getSession().getAttribute("SPRING_SECURITY_CONTEXT") != null) {
+            SecurityContext securityContext = (SecurityContext) req.getSession().getAttribute("SPRING_SECURITY_CONTEXT");
+            LOGGER.trace("Session for request {} contains security context. Username: {}; Permissions: {} ",
+                    req.getPathInfo(), securityContext.getAuthentication().getName(), securityContext.getAuthentication().getAuthorities());
+        }
```
pgesek: I would log something in else here

PR #217, opened 2015-11-23
MOTECH-1993: Entity columns widths are stored in the cookie.

tstalka:


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4792,6 +4787,26 @@
...
+            var cookieName = $scope.getTableWidthCookieName($scope.selectedEntity);
+            if ($.cookie(cookieName)) {
+               return JSON.parse($.cookie(cookieName));
+            }
+            return {};
+        };
+
+        $scope.getTableWidthCookieName = function(entity) {
+            var username = $rootScope.username || '';
+            return username + 'table.width_org.motechproject.mds.databrowser.fields.' + entity.className + '#' + entity.id;
```
pgesek: add a dot between table and username

pgesek: do we need both the classname and id?

PR #222, opened 2015-11-24
MOTECH-1990: Tasks Module> MDS Data Source > Incorrect Format for Key

tstalka:


modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/KeyInformation.java
```diff
@@ -110,7 +110,7 @@ public static KeyInformation parse(String input) {
         if (prefix.equalsIgnoreCase(TRIGGER_PREFIX)) {
             key = new KeyInformation(input, prefix, withoutManipulation, manipulations);
         } else if (prefix.equalsIgnoreCase(ADDITIONAL_DATA_PREFIX)) {
-            Pattern pattern = Pattern.compile("([a-zA-Z0-9]+)\\.([\\.a-zA-Z0-9\\-]+)#(\\d+)\\.(.+)");
+            Pattern pattern = Pattern.compile("([a-zA-Z0-9]+)\\.([\\.a-zA-Z0-9\\-_]+)#(\\d+)\\.(.+)");
```
pgesek: We should extend KeyInformationTest to catch this case

PR #228, opened 2015-11-27
MOTECH-1994: Saving task without task action does not cause task to be uneditable

LukSkarDev:
* Saving task without task action will not cause this task to be uneditable. Instead user will be informed
about problems to solve before task can be enabled. List of validation errors appears every time user saves task that can't be enabled.
* Channel action dropdown list will not appear until the channel is selected.


sebbrudzinski:
test this please

pgesek:
test this please

modules/tasks/tasks/src/main/resources/webapp/messages/messages.properties
```diff
@@ -17,6 +17,7 @@ task.confirm.changeObject=Changing data source object could invalidate form. Con
 task.success.removed=Task removed
 task.success.saved=Task saved
 task.success.savedAndEnabled=Task saved and enabled
+task.success.savedWithWarning=This task is saved but can't be enabled until following errors won't be resolved:
```
pgesek: was saved

pgesek: , but

pgesek: will apostrophes work like this?

pgesek: The task

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskService.java
```diff
@@ -17,11 +19,13 @@
...
     /**
-     * Saves the given task in the database.
+     * Saves the given task in the database and returns the list
+     * of errors that will occur when enabling task.
      *
      * @param task  the task to be saved, not null
+     * @return list of task errors
      */
-    void save(final Task task);
+    Set<TaskError> save(final Task task);
```
pgesek: We should also improve the error alert that is displayed when clicking enable in the task list view. Currently it doesn't list the errors, just a cryptic 'can't do this' message.

modules/tasks/tasks/src/main/resources/webapp/partials/form.html
```diff
@@ -386,8 +388,10 @@ <h3 class="popover-title">{{msg('task.tooltip.availableTriggers')}}</h3>
                                                 </li>
                                             </ul>
                                         </div>
-                                        <span class="offset-4">{{msg('task.action')}}:</span>
-                                        <div class="btn-group offset-15">
+
+
+                                        <span ng-show="selectedActionChannel[$index]" class="offset-4">{{msg('task.action')}}:</span>
+                                        <div ng-show="selectedActionChannel[$index]" class="btn-group offset-15">
```
pgesek: Something wrong here, I do not see any fields for actions

PR #233, opened 2015-12-01
MOTECH-2022 Consolidate static files

nickdotreid:
Created a stand alone directory for static resources called static and included it into the motech-platform-server. Additionally I added the start of files for a 'modern' frontend development pipeline.

added static resource path to .pom

added NPM and bower config files


nickdotreid:
Please test this

This reorganization does cause some minor bugs from images missing (including the favicons). This can be cleaned up in future work, but I feel like this is ready to be merged in or talked about a little.

I split up the commits into logical chunks, I'm happy to squish them together more if that makes more sense.

One difficulty I have is getting an absolute path to the 'static/commons' directory where common resources are stored. If I were able to figure out this path, I'd be able to generate header files in gulp, and include them on to individual pages (which I think is ideal). The crux of this issue is that the platform-server bundle (bootstrap) is one level deep (/bootstrap/) while the server-bundle is two levels deep (/server/module/), so both files can't share the same header include.

Thoughts: @larubbio @sebbrudzinski @pgesek

pgesek:
@nickdotreid few comments/questions:

1) This cannot be merged, you have to merge your branch with the latest master
2) The build failed, due to an error about the Maven version being too old. Our Docker containers use apt-get install maven for installing Maven, so I'm guessing this plugin is a bit more bleeding edge then the package repositories. Any chance we could bring the version down? If not, then we will have to update dockers and the documentation to mention that Maven 3.1 is required.
3) Regarding the previous point, from what I can see this plugin installs node and bower. Is this behavior controlled somehow? What if have node/bower already installed?
4) I can see that we are using Gulp for uglification and all that good stuff. Is it still possible to use non minified files during development?
5) Regarding the context path, won't the jsp syntax help with this? You can use request.getContextPath()/getPathInfo() or something along those lines to figure it out. That being said, why not just rename /bootstrap to /server/bootstrap and be done with it?

nickdotreid:
@pgesek - (1) I'll merge in the master branch when I'm done making the other edits...

(2 & 3) Yes, I changed my Maven version to 3.1 (I think the repo supported version is 2.8) ~ Is supporting node & npm packages easier than upgrading Maven? Currently I'm using a Maven plugin that installs node/npm (which I'm happy to drop and replace with mojo executions in Maven)

^ I'll move forward with this unless I hear otherwise

(4) I've been flipping back and forth myself... would it make sense to change the header type depending on an environment called 'DEVELOPMENT' ? (The header would be swapped at runtime in Tomcat)

Another option would be to compile the non-uglified version and call it the same file name as the uglified version, but that just sounds gross and I'd like to allow for resource paths to be swapped out

Do you know where/what runs the 'quick' resource development in the OSGI modules? I think following that model makes sense

(5) When I used request.getContextPath() I'd get different values/depths depending on which module I was in — I like changing /bootstrap to /server/bootstrap and hardcoding everything to ../../ (its not pretty, but it seems like the quickest solution)

pgesek:
@nickdotreid 

(2 & 3) I'm ok with that. We wil have to update the docs and dockers still. Can you let me know what exactly needs to be set up? I'm guessing something along the lines of:

apt-get install node
npm install bower

4) Yes using the DEVELOPMENT var makes sense. In Motech, org.motechproject.osgi.web.ext.FileSystemAwareUIHttpContext is responsible for that. The check for development environment is made in org.motechproject.osgi.web.ext.HttpContextFactory.

5) Sounds fine

nickdotreid:
@pgesek (2&3) All that needs to be installed in NodeJS and npm

apt-get install nodejs npm

packages such as bower and gulp are installed by npm when the motechproject server is installed.

nickdotreid:
@pgesek -- I'm lost here...

I created an Interceptor to define variables such as 'staticPath' and 'developmentMode' to use in the JSP files. I created the files in the motech server war, which I thought would make it accessible to the server-bundle as well — but it compiles and then I get a 'class doesn't exist' exception at runtime

I tried moving the logic into osgi-web-utils — but then the server war can't use the files since the OSGI isn't started... even though this will also compile just fine. I wanted to move the logic here because this is where the aware file system and existing environment variable related code lives.

What is the best approach to take for creating a reusable set of UI related classes? Should I create a new set of classes somewhere?

I've tried to figure out where I could set an interceptor for the entire stack, but am a little confused about which web-app definition file is used by the proxy module

pgesek:
@nickdotreid Generally speaking the war code shouldn't be reused with OSGi modules. The war should be kept minimal. The reason are classloader differences, those classes wont be able to interact, although there are ways around that. 

If you place the classes inside a platform bundle, the war should be able to use them since they will be on its classpath. OSGi starting shouldn't really have anything to do with it, since the classes that the war will be using will be different from the classes that the OSGi bundle uses - if you try mixing them up you would get the baffling org.example.Class cannot be cast to org.example.Class error.

That being said, mixing the war and the bundles will be a mess and classloader hell, best to be avoided. We always kept the war minimal, the only exception was adding the bootstrap page for the initial config.

nickdotreid:
Good to know... I'll give it a quick little go and see if I have any
luck... Then I guess I'll settle for a little technical debt as it seems
unavoidable

nickdotreid:
@pgesek - This branch is ready for merging... I'm not sure how to add nodejs and npm to the CI ~ but I'd be happy to give it a go if you can give me a clue about where to look...

I did something a little weird (that I kinda like) -- instead of trying to use the same static paths on both the war and the bundles, I created a gulp process that renders the same header using different static resource paths...

This goes into the odd bucket, because the jsp pages include a 'header.jsp' file that doesn't exist until maven builds the package

pgesek:
@nickdotreid I've installed node and npm in our docker containers:

jenkins@0a1fca136385:~$ npm -v
1.1.4
jenkins@0a1fca136385:~$ node -v
v0.6.12

pgesek:
test this please

pgesek:
@nickdotreid the build due to missing gulp, just giving a quick glance it seems that we execute the install commands in the server module, which is too late - server bundle will run first.

nickdotreid:
hrm — I thought it loaded the otherway around...

Here are a couple thoughts about how to fix it, I'd love to know your
thoughts
(1) Duplicate install code into server bundle module. If the install code
is run for a second time it won't download the files for a second time. I
don't LOVE this because we have the configuration commands in two different
places
(2) Create a set of maven tasks in the UI directory and add that as a
dependency for both modules

nickdotreid:
test this please

nickdotreid:
@pgesek - I'm a little confused why this build is failing, and would love some thoughts about how/where do go about debugging

I looked at the log message — the build fails because npm can't download a package (specifically called gulp-rename)

I can't reproduce this error locally... but I'm not running the build in docker (which could be the problem)

--> Is it possible that security restrictions or something could be preventing NPM packages from being downloaded?

pgesek:
@nickdotreid unlikely

 http://stackoverflow.com/a/21715730

Eh, looking at this answer, seems like I installed pre cold war versions of npm and nodejs. Might be because our Image of Docker is based of Ubuntu 12 I think. I will replace it with a manual install of the latest version tomorrow.

pgesek:
@nickdotreid I've updated the versions

$ npm -v
3.3.12
$ node -v
v5.3.0

pgesek:
test this please

nickdotreid:
test this please

nickdotreid:
@pgesek I figured out the problem (I hope)

gulp and bower need to be installed globally (at least the way I'm using them)... adding the following to the docker container should do the trick...

npm install -g gulp bower

I'm going to update the documentation... and if I knew how to test the CI I'd try to make the changes... but...

pgesek:
@nickdotreid I've installed gulp and bower:

$ gulp -v
[10:26:52] CLI version 3.9.0
[10:26:52] Local version 3.9.0
$ bower -v
1.7.2

pgesek:
test this please

pgesek:
@nickdotreid FYI in order to update the CI slaves you have to:

1) Update the images in the docker-ci repo
2) Rebuild the images
3) Push the updated containers to Docker Hub
4) Ssh to nufu.motechproject.org and pull the new images
5) Restart the slaves

Currently I am doing this manually, but I am considering automating this to some extent.

pgesek:
@nickdotreid I've left you 1 comment regarding the environment check. In general everything seemed to work fine. Fix that one issue and I will merge this change

nickdotreid:
@pgesek Fixed...

PR #237, opened 2015-12-07
MOTECH-1988 ActiveMQ channels are started after received event

martokarski:


platform/event/src/main/java/org/motechproject/event/listener/EventConsumerStarter.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.motechproject.server.osgi.event.OsgiEventProxy;
+import org.motechproject.server.osgi.util.PlatformConstants;
+import org.osgi.service.event.Event;
+import org.osgi.service.event.EventHandler;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.integration.MessageChannel;
+import org.springframework.integration.message.GenericMessage;
+
+public class EventConsumerStarter implements EventHandler {
```
pgesek: Add javadoc

sebbrudzinski:
WARNING: EventAdmin: Blacklisting ServiceReference [[org.osgi.service.event.Even
tHandler] | Bundle(org.motechproject.motech-platform-event [113])] due to timeou
t!

Can you verify? It seems that some modules keep waiting for this and won't start until this timeout takes place.

platform/event/src/main/java/org/motechproject/event/listener/EventConsumerStarter.java
```diff
@@ -0,0 +1,42 @@
...
+import org.osgi.service.event.EventHandler;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.integration.MessageChannel;
+import org.springframework.integration.message.GenericMessage;
+
+/**
+ * Handles incoming events and starts ActiveMQ outbound channels.
+ */
+public class EventConsumerStarter implements EventHandler {
+    private MessageChannel controlChannel;
```
pgesek: add INFO level loggin to this class

sebbrudzinski:
Doesn't seem to work properly. I had 1400 messages pending and they got dequeued right after I started the server. Please verify.

martokarski:
test this please

sebbrudzinski:
Found some more issues:

1. When I installed all modules (main repo + modules repo) the channels seem to never start. It works when only having main repo installed.
2. It seems like it's not possible to send any MotechEvent while channels are not started. This should not be a case. I should be able to send an event, but their delivery should start after all modules have started. 
3. My logs are flooded with exceptions:
[event_changes_exceptions.txt](https://github.com/motech/motech/files/78607/event_changes_exceptions.txt)

martokarski:
test this please

PR #240, opened 2015-12-09
MOTECH-2018: Module restarted after failing

tstalka:
Now when refreshing of the module will fail then MDS will automatically restart it.
Style of the Mds settings page was changed in this pull request (now it looks like settings from the Admin module).


tstalka:
test this please

PR #242, opened 2015-12-10
MOTECH-2026: Added a 5m wait for app ctx in listeners

pgesek:
Trash / history listeners will now wait for 5 minutes
for the application context if it is not available
once they are called. This should resolve the issue
with a null pointer exception thrown, due to listeners
executing too early in rare cases.


sebbrudzinski:
Doesn't fix the issue from the ticket

sebbrudzinski:
@pgesek OK to merge

PR #243, opened 2015-12-11
1926

atish160384:
Removed duplicate datepicker code from modules and added common code in server-bundle.


pgesek:
Conflicts need to be resolved in order for this to get merged

kmadej:
You can't remove mdsDatePicker. This picker is used 8 times in mds-web module. If possible try to use in mds-web main datePicker, however remember to check if it works correctly.

Additionally in modules repo try to reuse existing datePicker, moving datePickers(for example commcareDatePicker) from module repo to motech repo is not the best solution.

PR #249, opened 2015-12-17
MOTECH-2028: Made unique editable from schema editor UI

pgesek:
* Only non DDE fields
* Only non collection types

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
@@ -20,15 +20,15 @@ import org.motechproject.mds.domain.Field;
 import org.motechproject.mds.domain.Type;
 import org.motechproject.mds.enhancer.MdsJDOEnhancer;
 import org.motechproject.mds.ex.entity.EntityCreationException;
+import org.motechproject.mds.helper.ClassTableName;
 import org.motechproject.mds.helper.EntitySorter;
-import org.motechproject.mds.util.JavassistUtil;
 import org.motechproject.mds.javassist.JavassistLoader;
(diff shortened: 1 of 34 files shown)


pgesek:
test this psql

pgesek:
test this psql

pgesek:
test this psql

pgesek:
test this psql

platform/mds/mds/src/main/resources/db/migration/default/V45__MOTECH-2028.sql
```diff
@@ -0,0 +1,6 @@
+-- add unique column in Entity table --
+
+ALTER TABLE "Field" ADD "uniqueField" boolean NOT NULL DEFAULT FALSE;
+
```
tstalka: add new fields to the EntityDraft

tstalka:
What if we already have the data in the entity?

pgesek:
If it violates the new constraint, then the schema update will fail. We can add fancy validation later on, for now this is sufficient.

tstalka:
Add new Entity, add field , mark as unique, unmark and save changes

[ex.txt](https://github.com/motech/motech/files/66437/ex.txt)

pgesek:
test this psql

pgesek:
test this psql

pgesek:
test this please

platform/mds/mds/src/main/java/org/motechproject/mds/builder/impl/MDSConstructorImpl.java
```diff
@@ -304,6 +307,34 @@ public void updateFields(Long entityId, Map<String, String> fieldNameChanges) {
         }
     }
 
+    @Override
+    @Transactional
```
tstalka: @Transactional("dataTransactionManager")

PR #257, opened 2015-12-22
MOTECH-2061: Changed the Flyway placeholder prefix

pgesek:
Changed from the default ${ to $flyway{ . Resolves issues with Velocity
strings in migrations getting parsed.


pgesek:
test this 0.25.X

PR #265, opened 2016-01-04
MOTECH- 1732: Disable DetachAllOnCommit

tstalka:


kmadej:
you can follow the issue here : https://github.com/motech/motech/pull/470

PR #266, opened 2016-01-04
MOTECH-2039: Manual task retry

LukSkarDev:
This commit adds possibility to manual trigger execution of failed task. The "retry" button is available in task history for failed task executions.


modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/TaskActivity.java
```diff
@@ -41,9 +42,12 @@
     private TaskActivityType activityType;
 
     @Field(displayName = "StackTrace element")
-    @Column(length = 8096)
+    @Column(length = 16192)
```
pgesek: Can we change this to a text/clob column instead of adding and adding here?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/web/ActivityController.java
```diff
@@ -93,4 +99,16 @@ public long getTaskActivityCount(@PathVariable Long taskId, @PathVariable String
...
+
+    /**
+     * Retries task execution for activity with the given ID.
+     *
+     * @param activityId the ID of activity for which task should be retried
+     */
+    @RequestMapping(value = "/activity/retry/{activityId}", method = RequestMethod.POST)
+    @ResponseStatus(HttpStatus.OK)
+    public void retryTask(@PathVariable Long activityId) {
+        TaskActivity activity = activityService.getTaskActivityById(activityId);
```
pgesek: Can we move this logic to the task trigger handler and make it one transaction there?

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1461,6 +1460,11 @@
                  });
             });
         };
+
+        $scope.retryTask = function (activityId) {
+            motechAlert('task.retry.info', 'task.retry.header');
+            $http.post('../tasks/api/activity/retry/' + activityId);
```
pgesek: we should handle success and error for the http post somehow

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/TaskTriggerHandlerTest.java
```diff
@@ -1721,6 +1713,18 @@ private MotechEvent createEvent(boolean withCustomParser) {
         return map;
     }
 
+    private void prepareActivityServiceAnswer() {
+        doAnswer(new Answer<Void>() {
```
pgesek: Instead of doing this, create a new hashmap in handleError in TaskTriggerHandler. In general its better not to mix these maps

PR #268, opened 2016-01-07
MOTECH-2010: Reworked UI for Scheduler

ngraczewski:
Scheduler UI now looks similar to the Tasks UI. Added Pause/Resume and Delete features
for jobs.


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -771,6 +779,62 @@ public void unscheduleAllJobs(String jobIdPrefix) {
...
+        JobDetail detail = scheduler.getJobDetail(key);
+
+        if (detail == null) {
+            throw new MotechSchedulerException(String.format("Job doesn't exist:\n %s\n %s", key.getName(),
+                    key.getGroup()));
+        }
+
+        JobDataMap map = detail.getJobDataMap();
+
+        if (map != null && map.getBooleanValue("programmed")) {
```
kmadej: use const PROGRAMMED

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -771,6 +779,62 @@ public void unscheduleAllJobs(String jobIdPrefix) {
         return messageTimings;
     }
 
+    @Override
```
kmadej: the methods pauseJob, resumeJob, deleteJob look similar, can you put the repetitive code into one method ?

kmadej:
1. There is a problem with long names on smaller screen (with lower resolution). Try to handle this similar to Task UI.
2. The Status and Activity button are not the same as in reworked spec on JIRA. I'm not sure it is ok. Please ask decision-making person.
3. I created schedule runonce job using task :
a) name of job is 'mysubject-null-runonce',  what 'null' means, I think it shouldn't be in name
b) I don't have information about my job. I should have 'job type', 'start date'
4. Pause/Resume/Delete buttons work as expected.

kmadej:
needs rebase

PR #269, opened 2016-01-08
1659

atish160384:
* Database is migrated to V44
* Added Type validations for the new types
* Added form validations for the new types


atish160384:
Database will be migrated to V3

PR #270, opened 2016-01-08
MOTECH-1989 Add UI information about not started ActiveMQ

martokarski:


sebbrudzinski:
1. We should adjust the information to clearly mention that it is a status of the ActiveMQ inbound channel. Currently it might be confusing for the users who see "X" near ActiveMQ, even though it is started and running.
2. I killed ActiveMQ but it still shows tick, as everything was OK.

sebbrudzinski:
@pgesek LGTM

PR #271, opened 2016-01-08
MOTECH-2035: Fixes bug after schema regeneration in MDS

kmadej:
This commit fixes bug, when NucleusException was thrown after schema regeneration
in MDS.

In MDS we use MDS JDO annotation reader, which extends the regular
JDOAnnotationReader from datanucleus. This class was introduced because
regular reader would not read field annotations for metadata if there was
no class level JDO annotations. In this class we fake the PersistenceCapable
annotation.

It works as expected, however sometimes after schema regeneration
datanucleus read wrong Metadata for classes. This is connected with this reader.
When we are loading entities using MDSClassLoader our reader implementation is
needed to see field annotations, however after package.jdo file is ready,
there is no need to fake PersistenceCapable annotation. It's why I added the
condition with BundleClassLoader.

After this changes, the bug should be fixed.

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/jdo/MdsJdoAnnotationReader.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/jdo/MdsJdoAnnotationReader.java
@@ -1,6 +1,7 @@
 package org.motechproject.mds.jdo;
 
 import org.apache.commons.lang.ArrayUtils;
+import org.apache.felix.framework.BundleWiringImpl;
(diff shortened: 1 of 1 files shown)


kmadej:
For reviewer : 
It's not easy to reproduce this bug. However to me the easiest way was : 
1. Install MOTECH with module CSD
2. Make simple change in EmailRecord schema (it can be check/uncheck read-only property)
3. Save changes
4. When the bug didn't occurred I restarted MOTECH and repeated steps (After first schema regeneration was success I've never reproduced the bug in next tries. I recommend to restart MOTECH and repeat the steps

PR #272, opened 2016-01-08
MOTECH-2060: Made Flyway properties configurable

pgesek:
Flyway properties are now handled the same way that DN properties
are. We have separate settings for data and schema databases.
Configuration is handled by the core-config module.


pgesek:
test this please

sebbrudzinski:
@pgesek Looks good to me. We will also need a pull request on the modules repository that adds the clearing of the flyway properties in the maven-clean-plugin.

PR #277, opened 2016-01-15
MOTECH-2037: Make motech directory configurable

1stmateusz:
Motech directory now can be saved in bootstrap on startup


1stmateusz:
test this please

1stmateusz:
test this please

platform/config-core/src/main/java/org/motechproject/config/core/bootstrap/impl/BootstrapManagerImpl.java
```diff
@@ -62,8 +62,10 @@ public BootstrapConfig loadBootstrapConfig() {
 
     @Override
     public void saveBootstrapConfig(BootstrapConfig bootstrapConfig) {
+        configLocationFileStore.add(bootstrapConfig.getMotechDir());
```
pgesek: The config location is not ~/.motech , but ~/.motech/config

modules/admin/src/main/java/org/motechproject/admin/bundles/BundleDirectoryManager.java
```diff
@@ -30,6 +36,12 @@
      * @return the bundle directory
      */
     public String getBundleDir() {
+        if (configurationService != null) {
```
pgesek: configuration service cannot be null here

modules/admin/src/main/java/org/motechproject/admin/bundles/BundleDirectoryManager.java
```diff
@@ -30,6 +36,12 @@
...
+            BootstrapConfig bootstrapConfig = configurationService.loadBootstrapConfig();
+            if (bootstrapConfig != null) {
```
pgesek: can bootstrap config ever be null here?

modules/admin/src/main/java/org/motechproject/admin/bundles/BundleDirectoryManager.java
```diff
@@ -22,6 +25,9 @@
 @Component
 public class BundleDirectoryManager {
 
+    @Autowired
+    private ConfigurationService configurationService;
+
     @Value("${user.home}/.motech/bundles")
     private String bundleDir;
```
pgesek: It would make more sense to init this in postConstruct instead of during each read.

modules/testing-utils/testing-utils/src/main/java/org/motechproject/testing/utils/PollingHttpClient.java
```diff
@@ -169,7 +169,7 @@ public void setCookieStore(CookieStore cookieStore) {
     private <T> T executeWithWaitForUriAvailability(HttpUriRequest httpUriRequest, ResponseHandler<? extends T> responseHandler,
                                                     int expectedErrorCode) throws InterruptedException, IOException {
         final long timeoutInMillis = maxWaitPeriodInSeconds * 1000;
-
+
```
pgesek: remove the tab

platform/config-core/src/main/java/org/motechproject/config/core/bootstrap/impl/BootstrapManagerImpl.java
pgesek: Is this correct? I would rather expect adding this with /config appended, rather than doing during the file check

platform/config-core/src/main/java/org/motechproject/config/core/domain/BootstrapConfig.java
```diff
@@ -30,11 +30,14 @@
     public static final String CONFIG_SOURCE = "config.source";
     public static final String SQL_DRIVER = "sql.driver";
     public static final String OSGI_FRAMEWORK_STORAGE = "org.osgi.framework.storage";
+    public static final String MOTECH_DIR = "motech.dir";
     public static final String QUEUE_URL = "jms.broker.url";
 
-    public static final String DEFAULT_OSGI_FRAMEWORK_STORAGE = new File(System.getProperty("user.home"), ".motech"+File.separator+"felix-cache").getAbsolutePath();
+    public static final String DEFAULT_OSGI_FRAMEWORK_STORAGE = new File(System.getProperty("user.home"), ".motech" + File.separator + "felix-cache").getAbsolutePath();
+    public static final String DEFAULT_MOTECH_DIR = new File(System.getProperty("user.home"), ".motech").getAbsolutePath();
     private SQLDBConfig sqlConfig;
```
pgesek: can we have an empty line between 37 and 38?

platform/config-core/src/main/java/org/motechproject/config/core/filestore/ConfigLocationFileStore.java
```diff
@@ -34,12 +37,41 @@ public ConfigLocationFileStore(PropertiesConfiguration propertiesConfiguration)
...
      * @return the list of configuration locations
      */
-    public Iterable<ConfigLocation> getAll() {
-        List<String> configLocations = loadAll();
-
+    public Iterable<ConfigLocation> getAllConfigLocations() {
+        List<String> locations = getLocations();
+        List<String> configLocations = new ArrayList<>();
+        for (String location : locations) {
+            configLocations.add(new String(location + "/config/"));
```
pgesek: As mentioned, I would expect the config to be appended during the save of the location. Moreover, we should be a bit more clever about file separators, pretty sure this can go wrong on Windows or when the value of the location var ends with a slash already.

platform/config-core/src/main/java/org/motechproject/config/core/filestore/ConfigLocationFileStore.java
```diff
@@ -34,12 +37,41 @@ public ConfigLocationFileStore(PropertiesConfiguration propertiesConfiguration)
...
+     *
+     * @return the list of configuration locations
+     */
+    public Iterable<ConfigLocation> getAllMotechLocations() {
+        List<String> locations = getLocations();
+        return map(locations);
+    }
+
+    private List<String> getLocations() {
+        String defaultPath = new File(System.getProperty("user.home"), ".motech").getAbsolutePath();
```
pgesek: can you explain the purpose of this method?

platform/config-core/src/main/java/org/motechproject/config/core/service/CoreConfigurationService.java
```diff
@@ -74,6 +74,12 @@
     ConfigLocation getConfigLocation();
 
     /**
+     * Returns the Motech location where all Motech files are present.
+     * @return configLocation.
+     */
+    ConfigLocation getMotechLocation();
```
pgesek: Is this the motech directory? We shouldn't use the ConfigLocation object for this

platform/config-core/src/main/java/org/motechproject/config/core/service/impl/CoreConfigurationServiceImpl.java
```diff
@@ -101,14 +101,33 @@ public void saveBootstrapConfig(BootstrapConfig bootstrapConfig) {
 
     @Override
     public ConfigLocation getConfigLocation() {
-        Iterable<ConfigLocation> configLocations = configLocationFileStore.getAll();
+        return getLocation(true);
+    }
+
+    @Override
+    public ConfigLocation getMotechLocation() {
```
pgesek: We should keep the config location separate from the new directory location. The only thing tying them together should be the value of the default config location, based on the motech.dir, not user.home.

platform/config-core/src/main/resources/config-locations.properties
```diff
@@ -1 +1 @@
-config.location=${sys:user.home}/.motech/config/, /etc/motech/, \\${sys:user.home}/.motech/config/
```
pgesek: This should stay as it was. We should support ${motech.dir} in this file

testdata/config-locations.properties
```diff
@@ -1 +1 @@
-config.location=${sys:user.home}/.motech/configtest/, ${sys:user.home}/.motech/config/, /etc/motech/
\ No newline at end of file
+config.location=${sys:user.home}/.motech/configtest, ${sys:user.home}/.motech
```
pgesek: should stay as it was, with ${sys:user.home} getting eventually replaced with ${motech.dir}

platform/config-core/src/main/java/org/motechproject/config/core/service/impl/CoreConfigurationServiceImpl.java
```diff
@@ -101,14 +101,16 @@ public void saveBootstrapConfig(BootstrapConfig bootstrapConfig) {
 
     @Override
     public ConfigLocation getConfigLocation() {
-        Iterable<ConfigLocation> configLocations = configLocationFileStore.getAll();
+        Iterable<ConfigLocation> locations;
+        locations = configLocationFileStore.getAll();
```
pgesek: can be done on 1 line

platform/config-core/src/main/java/org/motechproject/config/core/service/impl/CoreConfigurationServiceImpl.java
```diff
@@ -101,14 +101,16 @@ public void saveBootstrapConfig(BootstrapConfig bootstrapConfig) {
...
         StringBuilder sb = new StringBuilder("");
-        for (ConfigLocation configLocation : configLocations) {
+        for (ConfigLocation configLocation : locations) {
             sb.append(configLocation.getLocation()).append(' ');
             Resource configLocationResource = configLocation.toResource();
             try {
-                Resource motechSettings = configLocationResource.createRelative(ConfigurationConstants.SETTINGS_FILE_NAME);
-                if (motechSettings.isReadable()) {
+                Resource motechSettings;
```
pgesek: same as above, why split into two statements?

platform/config-core/src/test/java/org/motechproject/config/core/domain/BootstrapConfigTest.java
```diff
@@ -16,23 +16,23 @@
...
     @Test(expected = MotechConfigurationException.class)
     public void shouldThrowExceptionIfDbConfigIsNull() {
-        new BootstrapConfig(null, ConfigSource.FILE, null, null);
+        new BootstrapConfig(null, ConfigSource.FILE, null, null, null);
     }
 
     @Test(expected = MotechConfigurationException.class)
     public void shouldThrowExceptionIfQueueUrlIsInvalid() {
-        new BootstrapConfig(new SQLDBConfig(sqlUrl, sqlDriver, null, null), ConfigSource.FILE, null, "invalid.url");
+        new BootstrapConfig(new SQLDBConfig(sqlUrl, sqlDriver, null, null), ConfigSource.FILE, null,null, "invalid.url");
```
pgesek: space after ,

platform/mds/mds/src/main/java/org/motechproject/mds/config/MdsConfig.java
```diff
@@ -145,7 +145,6 @@ public File getFlywayMigrationDirectory() {
         String flywayLocation = getFlywayLocations()[0];
         File migrationDirectory = new File(System.getProperty("user.home"), Constants.EntitiesMigration.MIGRATION_DIRECTORY);
```
pgesek: Is this correct, shouldn't we be using motech.dir here?

1stmateusz:
i didn't saw that autowired, it was indeed unnecessary, but in some mds integration tests coreConfigurationService is null at some point, so I added "if" which set default value of motech directory

pgesek:
We have a bean defined in mdsTestContext - mockCoreConfiguration service. It should be injected in those tests unless there is an issue with this getting called from somewhere else. Do you have a stracktrace for the failure? If not, remove it and run it on jenkins

PR #280, opened 2016-01-18
MOTECH-2080: Cleaned up commons-api

pgesek:
* Removed old tenant handling code
* Added missing javadocs and serialversionid


sebbrudzinski:
@pgesek OK to merge

PR #281, opened 2016-01-19
MOTECH-2041: Updated the RPM package for 1.0

tstalka:
Tested with CentOS 7


tstalka:
If you need build this on the Ubuntu then you must install RPM -  "sudo apt-get install rpm"

PR #282, opened 2016-01-19
MOTECH-2040 Improve task retry mechanism

kmadej:
This commit improves task retry mechanism. Now we have two additional properties
in Task in order to define custom task retry mechanism :
numberOfRetries and retryIntervalInMilliseconds

The new task fields are respected by task import/export. Moreover there are
visible on the UI when editing or creating a task. I added validation on client
and server side. Also the retry interval field indicates through a placeholder,
that it accepts seconds.

I changed and added unit tests. I will add javadocs and integration tests for
task, when my idea for this feature will be accepted.


kmadej:
test this please

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskActionExecutor.java
```diff
@@ -231,19 +239,47 @@ private boolean callActionServiceMethod(ActionEvent action, Map<String, Object>
...
                 Method method = service.getClass().getMethod(serviceMethod, methodHandler.getClasses());
 
-                try {
-                    method.invoke(service, methodHandler.getObjects());
-                } catch (IllegalAccessException | InvocationTargetException e) {
-                    throw new TaskHandlerException(
-                            ACTION, "task.error.serviceMethodInvokeError", e,
-                            serviceMethod, action.getServiceInterface()
-                    );
+                int numberOfRetries = task.getNumberOfRetries();
```
pgesek: All of this reeks of overcomplication. If we are extending the event anyway, why not just put the max retry number inside the event?

pgesek:
The solution is needlessly complex to use in my opinion. Why not just have the retry number in the event, which if set would override the actual retry count. Is it possible to do the same for intervals? 
We can still have a custom error handler logic, but we can keep the retry logic out of it.

Also this sort of solution will have to be touched on the mailing list, since we are currently stepping outside the bounds of the ticket by extending the event module.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskActionExecutor.java
```diff
@@ -231,19 +239,47 @@ private boolean callActionServiceMethod(ActionEvent action, Map<String, Object>
...
+                            } else {
+                                LOGGER.info("Max retry count = {} reached for {} in {}", numberOfRetries, serviceMethod,
+                                        action.getServiceInterface());
+                            }
+                            throw new TaskHandlerException(
+                                    ACTION, "task.error.serviceMethodInvokeError", e,
+                                    serviceMethod, action.getServiceInterface()
+                            );
+                        } else if (retryIntervalTime != 0) {
+                            Thread.sleep(retryIntervalTime);
```
pgesek: This is bad, since we don't know the interval - it can be hours. A solution to this was discussed on the mailing list - we were supposed to use the scheduler through an event triggered job.

PR #284, opened 2016-01-19
MOTECH-2104 IE9 Automatically Refreshing on login page

nickdotreid:
The login page in IE9 caused the UI to constantly refresh. Updating Angular.js to version 1.2.29 fixed the problem. I also changed variable names from 'const' to 'var' because 'const' isn't supported below IE11.


larubbio:
lgtm

PR #286, opened 2016-01-20
MOTECH-2108: Fixes integration tests which use AbstractTaskBundleIT

kmadej:
After adding the @Transactional annotation to the TaskTriggerHandler, we need
to add some import packages to motech-task-test-utils module. Now integration
tests which use AbstractTaskBundleIT will work correctly.

I've already checked Commcare ITs and Schedule-tracking IT


kmadej:
test this please

PR #288, opened 2016-01-21
MOTECH-2040: Improves task retry mechanism

kmadej:
This commit improves task retry mechanism. Now we have two additional properties
in Task in order to define task retry mechanism:
numberOfRetries, retryIntervalInMilliseconds

The new task fields are respected by task import/export. Moreover there are
visible on the UI when editing or creating a task. I added validation on client
and server side. Also the retry interval field indicates through a placeholder,
that it accepts seconds.

How does it work ?
1. First of all we can define numberOfRetries and retryIntervalInMilliseconds
for task.
2. When we define values and task will fail, proper motech event will be sent
to schedule job.
3. In Scheduler module I added handler which listens to this schedule job event.
4. Scheduler handler method schedule RepeatingSchedulableJob with earlier
defined values for repeatCount and repeatIntervalInSeconds
5. Scheduler fires an event with subject for retryTask mechanism.
6. Task module handles this event, when task will be success the unschedule
event will be fired.
7. When task will be deleted, the unschedule event will be fired.
8. When task will be disabled, nothing happens.
9. When we not define this values for task, a task will behave the same as
before my changes

I've already open PR for this ticket(https://github.com/motech/motech/pull/282), however this changes are totally different that previous one, that's why I decided to open another PR


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerListener.java
```diff
@@ -0,0 +1,68 @@
...
+import org.motechproject.scheduler.contract.RepeatingSchedulableJob;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.Map;
+
+/**
+ * This service is used to schedule/unschedule jobs through the motech event.
+ */
+@Service
```
sebbrudzinski: Look more like a component. Morever, I don't think this class should be exposed looking at its methods, yet it's placed in the service package. Can you move this class to a better place?

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerListener.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String REPEAT_COUNT = "repeatCount";
+    private static final String REPEAT_INTERVAL_TIME = "repeatIntervalInSeconds";
+    private static final String JOB_SUBJECT = "jobSubject";
+
+    private MotechSchedulerService schedulerService;
+
+    /**
+     * Handles the motech event scheduling new job.
+     *
+     * @param event the event to be handled
```
sebbrudzinski: can we mention the event contains job parameters?

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerListener.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String UNSCHEDULE_JOB = "unscheduleJob";
+
+    private static final String REPEAT_COUNT = "repeatCount";
+    private static final String REPEAT_INTERVAL_TIME = "repeatIntervalInSeconds";
+    private static final String JOB_SUBJECT = "jobSubject";
+
+    private MotechSchedulerService schedulerService;
+
+    /**
+     * Handles the motech event scheduling new job.
```
sebbrudzinski: Seems like it only schedules repeating job, can we update the javadoc to mention this?

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerListener.java
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @param event the event to be handled
+     */
+    @MotechListener(subjects = { SCHEDULE_JOB })
+    public void handleScheduleJobEvent(MotechEvent event) {
```
sebbrudzinski: seems like it only schedules repeating jobs, can method name reflect it?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/Task.java
```diff
@@ -230,6 +242,27 @@ public void setTaskConfig(TaskConfig taskConfig) {
...
 
+    public int getNumberOfRetries() {
+        return numberOfRetries;
+    }
+
+    public void setNumberOfRetries(int numberOfRetries) {
+        this.numberOfRetries = numberOfRetries;
+    }
+
+    @Ignore
```
sebbrudzinski: this is neither a getter nor setter, the annotation is not necessary

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskTriggerHandler.java
```diff
@@ -135,17 +151,37 @@ public void handle(MotechEvent event) throws TriggerNotFoundException {
     }
 
     @Override
+    public void handleRetry(MotechEvent event) {
+        LOGGER.info("Handling the motech event with subject: {}", event.getSubject());
```
sebbrudzinski: Perhaps it would be wise, to distinguish this log from the regular handle

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskTriggerHandler.java
```diff
@@ -135,17 +151,37 @@ public void handle(MotechEvent event) throws TriggerNotFoundException {
...
+    }
+
+    @Override
     @Transactional
     public void retryTask(Long activityId) {
         TaskActivity activity = activityService.getTaskActivityById(activityId);
         handleTask(taskService.getTask(activity.getTask()), activity.getParameters());
     }
 
+    private boolean isTaskRetriesAlreadyScheduled(Map<String, Object> eventParams) {
```
sebbrudzinski: TaskRetry

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TriggerHandler.java
```diff
@@ -14,13 +14,23 @@
     /**
      * Registers this handler to listen for events with the given subject. This handler will now act as a MotechEvent
      * listener for the given subject and will get called by the event system when an event with the given subject is
-     * fired.
+     * fired. By default handler is for task trigger.
```
sebbrudzinski: ??

modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskValidator.java
```diff
@@ -452,4 +454,15 @@ private static TaskError validateDateManipulation(String manipulation, Parameter
         return errors;
     }
 
+    private static Set<TaskError> validateRetryOnFailureValues(int numberOfRetries, int retryIntervalInMilliseconds) {
+        Set<TaskError> errors = new HashSet<>();
+
+        if (numberOfRetries < 0 || retryIntervalInMilliseconds < 0) {
+            TaskError error = new TaskError("task.validation.error.invalidRetryOnFailureValues");
```
sebbrudzinski: Couldn't find a mapping for this key

modules/tasks/tasks/src/main/resources/webapp/messages/messages.properties
```diff
@@ -5,6 +5,10 @@ task.name=Task Name
 task.active=Active
 task.paused=Paused
 task.channel.deregistered=Channel not registered
+task.retryTaskOnFailure=Retry task on failure
+task.numberOfRetries=Retry Count
+task.retryInterval=Retry Interval
+task.retryIntervalPlaceholder=Please set value in seconds
```
sebbrudzinski: the value

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskServiceImplTest.java
```diff
@@ -226,6 +242,8 @@ public void shouldSaveTaskWithEmptyActionInputFields() {
 
         taskService.save(task);
         verify(triggerHandler).registerHandlerFor(task.getTrigger().getEffectiveListenerSubject());
+        // When task has not set number of retries, it should not register handler for retries
+        verifyNoMoreInteractions(triggerHandler);
```
sebbrudzinski: this won't be executed since taskService.save throws exception. Can you refactor and put those checks in the finally block?

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskServiceImplTest.java
```diff
@@ -156,6 +156,22 @@ public void shouldNotSaveTaskWithoutAction() {
...
+
+        taskService.save(t);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void shouldNotSaveTaskWithInvalidRetryIntervalValue() {
+        Task t = new Task("name", trigger, asList(action));
+        t.setRetryIntervalInMilliseconds(-10);
+
+        taskService.save(t);
```
sebbrudzinski: Can you verify in the finally block, that no save was performed?

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskServiceImplTest.java
```diff
@@ -156,6 +156,22 @@ public void shouldNotSaveTaskWithoutAction() {
     }
 
     @Test(expected = ValidationException.class)
+    public void shouldNotSaveTaskWithInvalidRetryNumberValue() {
+        Task t = new Task("name", trigger, asList(action));
+        t.setNumberOfRetries(-3);
+
+        taskService.save(t);
```
sebbrudzinski: Can you verify in the finally block, that no save was performed?

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/TaskTriggerHandlerTest.java
```diff
@@ -1514,6 +1534,175 @@ public void shouldHandleTriggerWithCustomParser() throws Exception {
...
+        handler.handleRetry(event);
+        ArgumentCaptor<MotechEvent> captorEvent = ArgumentCaptor.forClass(MotechEvent.class);
+
+        assertEquals(1, task.getFailuresInRow());
+        // since we already scheduled task retries, we should not send once again schedule job event
+        verify(eventRelay).sendEventMessage(any(MotechEvent.class));
+    }
+
+    @Test
+    public void shouldNotScheduleTaskRetries() throws Exception {
```
sebbrudzinski: Condition is missing in the test name

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskServiceImplTest.java
kmadej: in this test taskService.save doesn't throw exception

sebbrudzinski: That's right, my bad. Sorry!

kmadej: No problem !

sebbrudzinski:
@pgesek  The code looks good to me.

PR #290, opened 2016-01-21
MOTECH-2096 MDS Data Browser: display issues with enums and lists

wstrzelczyk:


pgesek:
I am unable to select a value in the selects now.

Repro:

1) Go to MDS
2) Click Add next 'Notification Rule'
3) Change the value for the select inputs

Result:
The value for the select goes to unselected, and I am unable to change it.

PR #294, opened 2016-01-22
MOTECH-2097: Fixes REST Api Documentation

kmadej:
This commit fixes REST Api Documentation. I had to add 2 javascript files from
swagger-js library to motech.js. Also I changed a swagger version to the old
one - 2.1.1-M1.


pgesek:
I think the style for the landing page is off

PR #296, opened 2016-01-25
MOTECH-2103: Improves grid selection fields mechanism

kmadej:
First of all, I changed collection type used in UserPreference for
selected/unselected fields. Now we use Set instead of List, which means
that we will not see duplicated rows in this tables.
Secondly, I have overridden hashCode() and equals() methods for Field class.
Now it should always properly remove Field objects from collection.

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/domain/Field.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/domain/Field.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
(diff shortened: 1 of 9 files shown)


platform/mds/mds/src/main/java/org/motechproject/mds/domain/Field.java
```diff
@@ -610,4 +611,25 @@ public boolean isUiChanged() {
     public void setUiChanged(boolean uiChanged) {
         this.uiChanged = uiChanged;
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name);
```
pgesek: Doesn't Datanucleus advise against using id in hash?

sebbrudzinski:
test this psql

kmadej:
test this psql

PR #298, opened 2016-01-27
MOTECH-2099: Removed Delete role buton while adding new role

sebastianceronik:
Delete role button was available because addOrEdit variable wasnt checked in few divs in html.
Added checking for it.


platform/web-security/src/main/resources/webapp/partials/role.html
```diff
@@ -87,22 +87,22 @@
                 <input ng-show="role.deletable" ng-disabled="roleForm.$invalid" class="btn btn-primary" type="submit" value="{{msg('security.save')}}" ng-click="saveRole()"/>
                 <input class="btn btn-default" type="submit" value="{{msg('security.cancel')}}" ng-click="cancelRole()"/>
             </div>
-            <div ng-show="role.deletable" class="form-group">
+            <div ng-show="role.deletable && addOrEdit =='edit'" class="form-group">
```
sebbrudzinski: Can this String take any other values than "add" or "edit"? Would it make sense to refactor this to a flag, for example "isEdit"?

PR #299, opened 2016-01-27
MOTECH-1412: Flyway migrations can run after MDS generates entity schema

sebbrudzinski:
The so called "module migrations" allow modules to define migrations, that run after MDS generates tables backing up created entities. The modules should expose their migrations in their resource directory, under db/migrations/(default/mysql). Those migrations will be copied to the .motech directory and ran after MDS generates schema.


sebbrudzinski:
test this 0.25

sebbrudzinski:
test this 0.25

sebbrudzinski:
tets this 0.25

sebbrudzinski:
test this 0.25

PR #308, opened 2016-02-01
MOTECH-2057: Prevent saving "null" String in settings

pgesek:
* Check for null
* Added a contructor to SettingsOptions (why should we need to use a
map entry?)


kmadej:
Looks good to me

PR #315, opened 2016-02-02
Fixed problem with loading setting page

wstrzelczyk:


sebbrudzinski:
@nickdotreid I've merged Waldek's fix, that adjusts your change for IE9. The reason is that the changes broke the startup and login page (the UI would hang after the initial module loading). This has been fixed by reverting the changes of Angular version, but I'm assuming that now the loading on IE9 doesn't work again, so you might want to look into that.

nickdotreid:
Cool, ill take a peek and see what sense I can make of it....

PR #319, opened 2016-02-03
MOTECH-2093: Changed mds-web/pom.xml to export only necessary packages

sebastianceronik:
MDS Web was exporting all packages contained in it. added <Export-Package> to export only web.service and web.domain used by ebodac.


pgesek:
Rename the package mds.web.ex to mds.web.exception for consistency

PR #322, opened 2016-02-03
MOTECH-2133: DatePicker - 'Now' button is shown like inactive

1stmateusz:


pgesek:
The fix version for this was MOTECH-1.0. The fact that it was discovered on 0.26.X does not mean the fix is required on 0.26.X. We do not plan any more 0.26.X releases in the near future since Ebodac is moving to 0.27, I am closing this .

PR #323, opened 2016-02-03
MOTECH-2133: 'Now' button is shown like inactive

1stmateusz:


platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -502,6 +502,9 @@
             link: function (scope, element, attr, ngModel) {
                 var isReadOnly = scope.$eval(attr.ngReadonly);
                 if(!isReadOnly) {
+                    angular.element(element).click( function() {
```
pgesek: Is this a good solution? We switch classes onclick? Isn't there a way of doing this without js? I would imagine the button isn't supposed to be greyed out by default. 

The old class returns after I click the 'Now' btw.

1stmateusz: The onclick function is assigned to input, not to "now" button, so after clicking on input the function search for that button and it does not switch classes but removes the "secondary" and add "primary" one. The "secondary" class is assigned to "Now" button by default in JQuery.

pgesek: For me, switching is the same as removing one class and adding a different one, but maybe my definition is wrong? I still don't get why we need to use javascript. Won't overriding the class 'ui-priority-secondary' be a better solution? Doing stuff like this in onclick just seems odd.

The current javascript solution doesn't work after I click the 'now' button.

Moreover, if this the default datepicker behavior, should we even consider this a bug?

pgesek:
Javascript/css aside, this is broken because the class changes back to the original one after I click the now button.

platform/mds/mds-web/src/main/resources/webapp/css/mds.css
```diff
@@ -462,4 +462,11 @@
 .mds .form-inline .form-group.inline,
 .mds .form-inline .form-group label {
     display: inline-block;
+}
+
+.ui-priority-secondary,
```
pgesek: We should move this to some global css file, so that this affects all datepickers, not only the mds ones.

PR #324, opened 2016-02-03
MOTECH-2133: 'Now' button is shown like inactive

1stmateusz:


pgesek:
We don't need this in 0.27.X

PR #327, opened 2016-02-05
MOTECH-1869: Make creating an index for a lookup optional

1stmateusz:


platform/mds/mds/src/test/java/org/motechproject/mds/builder/impl/EntityMetadataBuilderTest.java
```diff
@@ -19,6 +19,8 @@
 import org.motechproject.mds.builder.SampleWithIncrementStrategy;
 import org.motechproject.mds.domain.ClassData;
 import org.motechproject.mds.domain.EntityType;
+//import org.motechproject.mds.domain.Field;
```
pgesek: ?!

platform/mds/mds/src/main/java/org/motechproject/mds/dto/LookupDto.java
```diff
@@ -27,6 +27,7 @@
     private String methodName;
     private boolean referenced;
     private List<String> fieldsOrder;
+    private boolean indexRequired;
```
pgesek: lets mark this true by default

PR #328, opened 2016-02-05
MOTECH-2091 Move classes into new packages and change osgi export-packages

martokarski:


platform/osgi-web-util/pom.xml
```diff
@@ -80,10 +80,11 @@
                         <Export-Package>
                             org.motechproject.osgi.web;version=${project.version},
                             org.motechproject.osgi.web.exception;version=${project.version},
-                            org.motechproject.osgi.web.ext;version=${project.version},
+                            org.motechproject.osgi.web.extension;version=${project.version},
                             org.motechproject.osgi.web.domain;version=${project.version},
```
pgesek: order the exports alphabetically - domain should be the second entry

PR #330, opened 2016-02-08
MOTECH-1762 Customize the embedded MDS view in IVR

akarczewski:
Even when visibility of all buttons were false, there was an empty space between title and table caused by toolbox div. So I added a possibility to remove it.

### DIFF
--- a/docs/source/get_started/custom_ui.rst
+++ b/docs/source/get_started/custom_ui.rst
@@ -400,6 +400,12 @@ Those variables let you decide which buttons will be visible in the embedded MDS
 
 .. code-block:: javascript
 
+    $scope.showToolbox
+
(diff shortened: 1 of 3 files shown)


pgesek:
test this please

docs/source/get_started/custom_ui.rst
```diff
@@ -400,6 +400,12 @@ Those variables let you decide which buttons will be visible in the embedded MDS
 
 .. code-block:: javascript
 
+    $scope.showToolbox
```
kmadej: something wrong here, the description about showToolbox should be above this variable

kmadej:
related with https://github.com/motech/modules/pull/127

PR #345, opened 2016-02-15
MOTECH-2063: Created UI for job creation/update in the Scheduler module

ngraczewski:


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/SchedulableJobBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+import org.quartz.Trigger;
+
+import static org.motechproject.scheduler.constants.SchedulerConstants.CRON;
+import static org.motechproject.scheduler.constants.SchedulerConstants.EVENT_TYPE_KEY_NAME;
+import static org.motechproject.scheduler.constants.SchedulerConstants.REPEATING;
+import static org.motechproject.scheduler.constants.SchedulerConstants.REPEATING_PERIOD;
+import static org.motechproject.scheduler.constants.SchedulerConstants.RUN_ONCE;
+import static org.motechproject.scheduler.constants.SchedulerConstants.UI_DEFINED;
+
+public final class SchedulableJobBuilder {
```
pgesek: javadoc

modules/scheduler/scheduler/src/main/resources/webapp/js/controllers.js
```diff
@@ -118,5 +129,172 @@
...
+
+            if (job.startDate) {
+                job.startDate = $scope.parseToDateTime(job.startDate);
+            }
+
+            if (job.endDate) {
+                job.endDate = $scope.parseToDateTime(job.endDate);
+            }
+
+            if (job.days) {
```
pgesek: why check if this exists if we immediately overwrite it on line 240? shouldn't we be checking $scope.job.days here?

modules/scheduler/scheduler/src/main/resources/webapp/js/directives.js
```diff
@@ -5,6 +5,79 @@
 
     var directives = angular.module('scheduler.directives', []);
 
+    /* General date time picker that can work as either date, time or datetime picker */
+    directives.directive('dateTimePicker', function ($timeout) {
```
pgesek: don't we have reusable datepickers? and wasn't this removed from teh scheduler module a few weeks ago?

PR #346, opened 2016-02-15
MOTECH-2088: moved MotechListenerType and EventCustomerStarter to new…

sebastianceronik:
… packages for clearance.

### DIFF
--- a/platform/event/pom.xml
+++ b/platform/event/pom.xml
@@ -128,7 +128,9 @@
                             org.motechproject.event;version=${project.version},
                             org.motechproject.event.listener;version=${project.version},
                             org.motechproject.event.listener.annotations;version=${project.version},
-                            org.motechproject.event.messaging;version=${project.version}
+                            org.motechproject.event.listener.annotations.types;version=${project.version},
+                            org.motechproject.event.messaging;version=${project.version},
+                            org.motechproject.event.starter;version=${project.version}
                         </Export-Package>
                         <DynamicImport-Package>
                             *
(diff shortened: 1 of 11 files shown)


platform/event/pom.xml
```diff
@@ -128,7 +128,9 @@
                             org.motechproject.event;version=${project.version},
                             org.motechproject.event.listener;version=${project.version},
                             org.motechproject.event.listener.annotations;version=${project.version},
-                            org.motechproject.event.messaging;version=${project.version}
+                            org.motechproject.event.listener.annotations.types;version=${project.version},
+                            org.motechproject.event.messaging;version=${project.version},
+                            org.motechproject.event.starter;version=${project.version}
```
pgesek: Why are exporting this starter package?

platform/event/src/main/java/org/motechproject/event/listener/annotations/types/MotechListenerType.java
```diff
@@ -1,4 +1,4 @@
-package org.motechproject.event.listener.annotations;
+package org.motechproject.event.listener.annotations.types;
```
pgesek: what's the point of this new package? It just contains 1 class, and both the old one and the new one are still exported.

pgesek:
Do we need org.motechproject.event.messaging exported?

sebastianceronik:
We need to export org.motechproject.event.messaging since classes there are used as helper classes for IT tests, not all though . Removed MotechEventTransformer and MotechEventHeaderMapper

pgesek:
We can go with this for now

PR #349, opened 2016-02-16
MOTECH-2090 Remove server-api bundle

martokarski:
These classes were mainly used by motech-admin or motech-server bundle, so I move its to those packages.


platform/server-bundle/src/test/java/org/motechproject/server/web/controller/ModuleControllerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import java.io.IOException;
+
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+/**
+ * Created by marek on 16.02.16.
```
sebbrudzinski: When you get a chance to fix the tests, please also remove this javadoc.

platform/server-bundle/src/main/java/org/motechproject/server/ui/BundleService.java
```diff
@@ -0,0 +1,27 @@
+package org.motechproject.server.ui;
+
+import org.motechproject.server.web.dto.BundleIcon;
+
+/**
+ * The <code>BundleService</code> interface provides methods for getting icons from the bundles.
+ */
+public interface BundleService {
```
pgesek: Shouldn't this be called BundleIconService?

martokarski:
Must be checked with modules

martokarski:
None of modules use this bundle

pgesek:
Needs rebase

PR #351, opened 2016-02-17
MOTECH-2014 Create modal directive - added plugin bootstrap3-dialog

wstrzelczyk:


platform/mds/mds-web/src/main/resources/webapp/partials/dataBrowser.html
```diff
@@ -5,6 +5,8 @@
                 <input type="text" ng-model="searchText" class="form-control input-sm input-large-fluid search-query" placeholder="{{msg('mds.form.searchByModuleEntity')}}">
             </div>
             <div class="btn-group offset-05">
+                <div ng-click="showModal()" class="btn btn-default btn-sm">showModal</div>
```
pgesek: remove this

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3461,6 +3470,49 @@
             $('#importInstanceModal').modal('show');
         };
 
+        $scope.showModal2 = function () {
```
pgesek: remove this

ui/src/js/common.js
```diff
@@ -1,18 +1,33 @@
     /* Common functions */
 
-function motechAlert(msg, title, params, callback) {
+function motechAlert(msg, title, params, Callback) {
```
pgesek: why are we capitalizing Callbacks?

PR #352, opened 2016-02-17
MOTECH-2176: Archetype IT build fails

1stmateusz:


utils/archetypes/minimal-bundle-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -31,6 +31,12 @@
             <artifactId>motech-pax-it</artifactId>
             <version>${motech.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-jdbc</artifactId>
```
pgesek: we should fix the version at source, not add an exclusion here

PR #353, opened 2016-02-18
Motech 2182: bootstrap-fileUpload is deprecated. We should refresh plugin used for fileupload.

1stmateusz:


pgesek:
Needs update with latest master

ui/bower.json
```diff
@@ -55,7 +55,8 @@
     "jquery.alerts-1.1": "http://labs.abeautifulsite.net/archived/jquery-alerts/jquery.alerts-1.1.zip",
     "jquery-ui": "~1.11.4",
     "bootstrap-multiselect": "~0.9.13",
-    "angular-bootstrap": "0.6.0"
+    "angular-bootstrap": "0.6.0",
```
pgesek: just curious, is simply switching the lib enough? Is the usage exactly the same?

1stmateusz: now I fixed some html files and it works fine

platform/ui-functional-tests/src/test/java/org/motechproject/testing/uifunctionaltests/DataServicesUIFT.java
```diff
@@ -0,0 +1,35 @@
+package org.motechproject.testing.uifunctionaltests;
+
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.motechproject.uitest.TestBase;
+import org.motechproject.uitest.page.DataServicesPage;
+
+public class DataServicesUIFT extends TestBase {
```
pgesek: Unrelated changes, clean the PR up

nickdotreid:
Hey, I know its a little late to be commenting on this PR, but...

The amount of repeated boiler-plate markup bugs me, and I'd love to see us starting to use a more DRY approach to our markup...

Could we wrap the file upload functionality into a directive? This way we can reduce mark up and make changes in the future easier? <-- This is a more Angular approach, and will help get us into position to make more changes later... (and maybe stick this directive into motech/ui, because I'm pretty sure there are file upload inputs else where in the interface)

Less importantly (and totally willing to punt on this issue) —I get why we are using Jansy, but I feel like pulling in a large library for a single component is overkill (unless we start using their other components // in which case their sidebar and alert components could be immedately useful)

I'd be happier seeing us use smaller more focused libraries, rather than adding frameworks for a single feature... and I'd highly recommend [ng-file-upload](https://github.com/danialfarid/ng-file-upload)

1stmateusz:
I had a lot of problems with adding ng-file-upload plugin in all places in motech, so I've decided to make a directive with jasny-bootstrap file-input plugin, bacause it has similar structure as our previous plugin.

pgesek:
@nickdotreid @1stmateusz  i see that you can customize what you download on jasny's page: http://www.jasny.net/bootstrap/customize/ (can we use bower/gulp for that?). Can't we just grab the fileinput component in the case if we deem including the whole thing overkill? 

I would also like to avoid including too much stuff as libraries, although we probably don't want to waste too much time on this.

nickdotreid:
@1stmateusz -- yea, screw ng-file-upload if its not fitting in well... I'll guess its because our API isn't 'angular' enough (and that's a different problem)

Let's not directly add jansy files to the repo, can you install jansy via bower, and then override the bower.json to include only the file-input components? (if you look at [ui/bower.json,](https://github.com/motech/motech/blob/master/ui/bower.json) you will see examples of overrides)

@pgesek Since jansy file-input is wrapped in a directive, that will make it easy to drop if we need/want to, and we won't be including the full stack in our compiled css/js...

Just BTW: I personally feel like getting the UI to a place where we can make global changes easily trumps getting the stack perfect.... and as you said, lets not waste too much time

1stmateusz:
@nickdotreid thanks for your advice, I use the override in ui/bower.json and it worked :)

pgesek:
I've tried installing this and I the upload inputs don't show on the UI at all. Is this in a working state?

1stmateusz:
Actually jasny-bootstrap file-upload plugin is not working properly on Firefox. I need to find other plugin and use it.

pgesek:
@1stmateusz ??? does jasny bootstrap not support firefox at all? I find this hard to believe to be honest

1stmateusz:
I moved file upload directive to every module where it's used. It should be fixed in MOTECH-2265 (same issue with period modal).

kmadej:
Now I can upload file on firefox and chrome. However I've got error while importing csv in data browser :
1. Go to Data Browser -> Notification rule
2. Create an instance
3. Export an instance to csv file
4. Delete this instance
5. Try to import csv 

Following error appears : 
screenshot from 2016-03-15 14 52 23

Nothing in logs. I've got similar problem while using import task in tasks module, so I assume this is connected with this ticket, not with the import csv feature.

platform/mds/mds-web/src/main/resources/webapp/partials/modals/import-entity.html
```diff
@@ -6,17 +6,11 @@
...
+                <form id="importInstanceForm" name="importInstanceForm" class="form-horizontal" action="{{getMdsEntityCsvImportPath(selectedEntity.id)}}" enctype="multipart/form-data" method="POST">
+                    <div class="form-group fileinput fileinput-new" data-provides="fileinput">
                         <label class="col-sm-3 control-label">{{msg('mds.import.file')}}</label>
                         <div class="form-inline col-sm-9">
-                            <div class="form-control input-auto uneditable-input input-xlarge fileupload-exists"><span class="glyphicon glyphicon-file"></span>&nbsp;<span class="mds-import fileupload-preview"></span></div>
-                            <span class="btn btn-default btn-file">
-                                <span class="fileupload-new">{{msg('mds.import.selectFile')}}</span>
-                                <span class="fileupload-exists">{{msg('mds.import.changeFile')}}</span>
-                                <input type="file" id="csvFile" name="csvFile" class="input-file" accept=".csv"/>
-                            </span>
```
kmadej: I see that we lost the feature - accepting only csv file. Is there any easy way to add this again ?

1stmateusz: Directive now can add accept attribute to file input.

pgesek:
No point in merging this, we should fix the issue with uglification first

kmadej:
I've found one usage of old fileupload in field-basic-defaultValue-blob. Please update also this one.

ui/src/js/directives.js
```diff
@@ -822,4 +822,32 @@
...
+                templateLoader = $http.get(url, {cache: $templateCache})
+                    .success(function (html) {
+                        tElement.html(html);
+                    });
+
+                return function (scope, element, attrs) {
+                    templateLoader.then(function () {
+                        element.html($compile(tElement.html())(scope));
+                        var input = element.find("input");
+                        if (attrs.accept !== null &&  attrs.accept !== "") {
```
kmadej: what about undefined ? the best way is to check it like the following : 'if (attrs.accept)'

kmadej:
We will merge this PR with old fileupload plugin in this few places. @1stmateusz can you update fileupload in field-basic-defaultValue-blob in another PR ?

Please revert this PR to version before my comment regarding old usage of fileupload in few places

kmadej:
Looks good to me. I checked on firefox and chrome: 
- import CSV button in data browser
- import task button in tasks
- install or update button in manage modules
- submit settings file in admin

PR #354, opened 2016-02-18
MOTECH-2183 Show columns in mds ui

martokarski:
Show columns name in empty table when initially none of them is selected


platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -1588,9 +1588,11 @@
                             if (noSelectedFields) {
                                 $('.page_' + target + '_center').hide();
                                 $('.ui-jqgrid-status-label').removeClass('hidden');
+                                $('#entityInstancesTable .ui-jqgrid-hdiv').hide();
```
pgesek: we should also handle trash/history grids. Do they have the same ID? If not, we should rather base this on a class perhaps

martokarski:
This show() and hide() calls were initially call only on create main table. All children of that div are set to hide, so I decided to remove this code.

platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -1358,12 +1358,10 @@
                                 }
                                 if ($('#instancesTable').getGridParam('records') > 0) {
                                     $('#pageInstancesTable_center').show();
-                                    $('#entityInstancesTable .ui-jqgrid-hdiv').show();
```
pgesek: what about trash and history? they still do this

martokarski:
test this please

PR #356, opened 2016-02-19
MOTECH-2129: fixed animation and redirect problem in IE9

sebastianceronik:
It turned out that IE9 lack support for css3 animations. Had to modify botstrapconfig.jsp.
I used  with 2 different gifs for circle and circle-spinner rather than using FontsAwesome with css3 animation.
Couldn't create class in css with src="" , because IE9 also cant handle content: src="" in css.


sebastianceronik:
changed gifs size since 20x20 was too big for its purpouse. ( AwesomeFonts uses 18x18 ) on bootstrap.

PR #359, opened 2016-02-22
MOTECH-2085 Cleanup server-config bundle

martokarski:


platform/server-config/pom.xml
```diff
@@ -76,12 +76,12 @@
                 <configuration>
                     <instructions>
                         <Export-Package>
+                            org.motechproject.config.domain;version=${project.version},
+                            org.motechproject.config.monitor;version=${project.version},
+                            org.motechproject.config.service;version=${project.version},
                             org.motechproject.server.config;version=${project.version},
```
pgesek: I feel this is a good occasion to get rid of the server.config and config packages - from what I recall this is purely legacy

martokarski:
Must be checked with modules

martokarski:
Changes in modules are made in https://github.com/motech/modules/pull/173

martokarski:
test this please

martokarski:
One IT test uses cms-lite module, which has dependency to server-config. This changes cause conflicts in cms-lite. Test will be uncomment after merging all changes in motech and modules.

kmadej:
After meeting with Pawel, we established that this test bundle used in admin IT should be change from cms-lite to tasks. @martokarski can you try change this bundle to tasks or other bundle from motech repo. If there will be any troubles, please write here.

kmadej:
Needs an update with latest master

platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -647,7 +648,7 @@ public void removeBundleRecords(List<ModulePropertiesRecord> records) {
      * @param settingsRecord  the settings to be add
      */
     @Transactional
-    public void addOrUpdateSettings(SettingsRecord settingsRecord) {
+    private void addOrUpdateSettings(SettingsRecord settingsRecord) {
```
kmadej: According to spring docs (http://docs.spring.io/spring/docs/3.0.x/spring-framework-reference/html/transaction.html#transaction-declarative-annotations) :

When using proxies, you should apply the @Transactional annotation only to methods with public visibility. If you do annotate protected, private or package-visible methods with the @Transactional annotation, no error is raised, but the annotated method does not exhibit the configured transactional settings. Consider the use of AspectJ (see below) if you need to annotate non-public methods.

In this method we need transaction, so it should be public.

pgesek:
test this please

kmadej:
Needs an update with latest master

pgesek:
1) BundlePropertiesService and SettingsService should get renamed to BundlePropertiesDataService and SettingsDataService, since they are data services. Moreover they should be moved to an .mds package.
2) Do we need the monitor package exported?

martokarski:
test this please

martokarski:
1) Renamed these classes and moved to the mds package.

2) Removed monitor package from exported packages. ConfigFileMonitor class was used in admin bundle code as bean, but this bean wasn't exported. De facto this field was always null (in SettingsServiceImpl from admin bundle), so I removed it.

PR #370, opened 2016-02-23
MOTECH-2142: fixed IE9 UI

sebastianceronik:
changed filter: progid:DXImageTransform.Microsoft.gradient input hexes to #AARRGGBB
previous #RRGGBB format was wrong ( see https://msdn.microsoft.com/pl-pl/library/ms532930(v=vs.85).aspx ) 

in IE9 for background colour to be correct we should use 6 digit hexes not 3 digit ones since IE9 cant interpret them correctly. for example #123 can be interpreted by IE9 as #000123 in some cases.

To make IE9 UI look less "HORRID". I created css class to override navbar-inverse wich was used only in IE9 for navigation bar colour. and removed gradients for mouse hover and select in ui to look just like chrome firefox. now UI should look the same on every supported browser.


sebastianceronik:
deleted bootstrap-theme.min.css ... we didnt need it in ui/src/css since this file is downloaded by bower from bootstrap.

sebastianceronik:
nowy obraz mapy bitowej

PR #376, opened 2016-02-25
MOTECH-2084: Cleaned up org.motechproject.tasks.domain package

ngraczewski:


pgesek:
has conflicts

modules/tasks/tasks/pom.xml
```diff
@@ -128,10 +128,14 @@
                         <Blueprint-Enabled>true</Blueprint-Enabled>
                         <Export-Package>
                             org.motechproject.tasks.service;version=${project.version},
+                            org.motechproject.tasks.service.osgi;version=${project.version},
```
pgesek: arrange the exports alphabetically

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/channel/ActionEvent.java
```diff
@@ -70,26 +68,13 @@ public ActionEvent(String name, String description, String displayName, String s
...
-        } else {
-            if (hasService() && info.hasService() && equalsService(info.getServiceInterface(), info.getServiceMethod())) {
-                result = true;
-            } else if (hasSubject() && info.hasSubject() && equalsSubject(info.getSubject())) {
-                result = true;
-            }
-        }
-
-        return result;
+    public ActionEvent(ActionEvent actionEvent) {
```
pgesek: why are we adding a copy constructor? seems the opposite of cleaning up

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -525,7 +527,7 @@ private TaskDataProvider findProviderByName(DataSource ds) {
             errors.addAll(taskValidator.validateAction(action, channel));
             TriggerEvent trigger = triggerEventService.getTrigger(task.getTrigger());
             Map<Long, TaskDataProvider> providers = getProviders(task);
-            ActionEvent actionEvent = channel.getAction(action);
+            ActionEvent actionEvent = ChannelUtils.getAction(channel, action);
```
pgesek: Is there a reason for moving this kind of methods to the utils? Simple methods like contains or getAction fit well with the domain classes in my opinion. If we move all logic to static helpers, that will not be very object oriented.

ngraczewski: This was done to solve circular dependencies. I couldn't think of anything more appropriate at that time.

sebbrudzinski: @ngraczewski I'd say you should rather refactor/create packages as necessary, rather than move the logic to the utils, when there's a circular dependency. @pgesek Can you answer/comment on that?

pgesek: not sure how channel.getAction creates a circular dependency. Doesn't the channel have the list of actions inside it anyway?

ngraczewski: Moving Task related classes to a separate package fixed the problem. The utils classes are gone.

PR #378, opened 2016-02-26
MOTECH-2092 Clean up the web-security API

sebbrudzinski:
* "ex" package renamed
* removed unnecessary exports
* added missing javadocs
* corrected some javadocs
* removed some unnecessary classes between
  data services and OSGi services
* test adjustments for the above


platform/web-security/pom.xml
```diff
@@ -160,41 +160,39 @@
...
                             org.motechproject.security.authentication;version=${project.version},
-                            org.motechproject.security.builder;version=${project.version},
                             org.motechproject.security.constants;version=${project.version},
                             org.motechproject.security.domain;version=${project.version},
-                            org.motechproject.security.ex;version=${project.version},
+                            org.motechproject.security.exception;version=${project.version},
                             org.motechproject.security.model;version=${project.version},
-                            org.motechproject.security.repository;version=${project.version},
                             org.motechproject.security.service;version=${project.version},
+                            org.motechproject.security.service.mds;version=${project.version},
```
pgesek: I believe we agreed a while ago on making the name of this package mds, not service.mds (service.mds is a bit redundant)

platform/web-security/src/main/resources/META-INF/motech/securityContext.xml
```diff
@@ -151,4 +151,10 @@
         <constructor-arg value="/module/server/changepassword"/>
     </bean>
 
+    <bean id="viewResolver" class="org.springframework.web.servlet.view.InternalResourceViewResolver">
```
pgesek: what's the reason for moving this? It's not related to the security configuration

sebbrudzinski: I wanted to move it away from osgi directory, since it's not OSGi related. I should have moved it to application context file though, my bad. Will fix.

platform/web-security/pom.xml
sebbrudzinski: OK. Will fix.

PR #391, opened 2016-03-03
MOTECH-2131: Replaced gradients created with image file 

pkornowski:
I swapped gradients with image file into a css gradients, also I removed unused image files ( which have been swapped ). I used sass mixin because of many different types of vendor prefixes. I added a gulp-sass, it is sass plugin for gulp, it is necessary because of need to compile .scss file.


nickdotreid:
If we are going to be integrating Sass, can we use [bourbon](http://bourbon.io/docs/#background) to generate the gradients? 

We will also need to add Ruby and 'gem install' Sass and Bourbon to the CI and installation documentation. Part of splitting the MOTECH UI from MOTECH was because the UI stack is getting more complicated...

@pgesek Do you think it makes sense to merge Ruby & Sass into the current build, even though we will be pulling it out of the MOTECH stack once the UI is complete?

I personally still want those background-images gone, and if this PR doesn't get into 0.28, I'm ok with making the gradients into solid colors until this is ported into the MOTECH-UI repo

pgesek:
@nickdotreid the fix version for this is 1.0, there are no plans for this getting into 0.28. Do we need this on 0.28? Why?

Can we install the ruby gem with maven? By the way, https://github.com/thoughtbot/bourbon here they mention that you can install with bower alternatively, do we need to get ruby involved?

If this requires ruby, we will have to update the ci containers. I'm guessing this is only a build dependency on ruby, not a runtime one?

nickdotreid:
There is no reason to include this in 0.28 — sorry if I got confusing....

I was under the impression to run Sass one needed ruby, but apparently I'm
completely wrong -- node-sass uses a C version of Sass -- and thats a
dependency of gulp-sass

Can we change the mixins to use bourbon linear-gradient instead of big ugly
vendor prefixed goo?

pgesek:
@nickdotreid Ok, so lets just install with bower then.  @pkornowski switch this to use bourbon, which should limit the boilerplate

pkornowski:
I installed bourbon from bower, but I tried and I couldn't get the path to it. I still gained errors, "File to import not found" when I wanted import bourbon into .scss file. When bourbon was pasted into css folder, it worked.  In Bourbon .scss I couldn't found 'filter' which is responsible for gradients in IE9. My question is: May I create separate .scss file with gradient mixin which would get as parameters gradient colors like in bourbon? We will be able to import that file and use that mixin, which should limit the line of code similar like the use of bourbon, also I can implement a filter IE9.

pgesek:
What do you mean 'couldn't get the path to it'. What path? From what I can see Bourbon supports IE 9, it should be fine, don't see a point in reinventing the wheel. If this an issue with the bower install, we can stomach the ruby dependency, hopefully maven can handle it?

nickdotreid:
Which version of bourbon is getting installed? You want node-bourbon, which uses node-sass
https://github.com/lacroixdesign/node-bourbon

A problem that you could be having is that bower installs front-end packages, which means they are not available to node/gulp

Let's try to avoid writing browser specific code when we don't need to — I'd personally rather have slight differences between browsers than litter our CSS with work arounds — and in this case, a background image won't make or break the "user's experience"

(but as justification for other bugs, an unreadable nav menu will)

@pkornowski feel free to email me directly if node-bourbon doesn't help anything (I should be online all day) (and sorry for the late reply)

pkornowski:
@nickdotreid Thank you for advice, that was helpful. I replaced my mixins with node-bourbon mixins, and I think that should be ok.

ui/package.json
```diff
@@ -12,6 +12,7 @@
     "bower": "~1.6.8",
     "gulp": "^3.9.0",
     "gulp-bless": "^3.0.1",
+    "gulp-sass": "2.3.0-beta.1",
```
pgesek: why are we using a beta version?

pkornowski:
I'm sorry about that, when I was looking for solution I replaced version 2.2.0 with beta version for a test. I forgot to change it back.

pkornowski:
Deleted white background in Modules -> Scheduler

PR #392, opened 2016-03-03
MOTECH-2086 Fixed MDS OSGi exports/packages/javadocs

sebbrudzinski:
* Added all exception classes to exports
 * Removed unnecesary exports and moved some classes
   to internal packages


platform/mds/mds/src/main/java/org/motechproject/mds/domain/EntityDraft.java
```diff
@@ -53,6 +53,10 @@
     @Element(column = "fieldName")
     private Set<String> uniqueIndexesToDrop = new HashSet<>();
 
+    public EntityDraft() {
+        fieldNameChanges = new HashMap<>();
```
pgesek: what's the point? fieldNameChanges has an initializer

platform/mds/mds/src/main/java/org/motechproject/mds/domain/EntityDraft.java
```diff
@@ -61,11 +65,6 @@ public void updateAdvancedSetting(AdvancedSettingsDto advancedSettings) {
         updateTracking(advancedSettings);
     }
 
-
-    public EntityDraft() {
-        fieldNameChanges = new HashMap<>();
```
pgesek: probably better to remove this altogether

platform/mds/mds/src/main/java/org/motechproject/mds/service/MotechDataService.java
```diff
@@ -241,7 +243,7 @@
      *
      * @param queryExecution implementation of the {@link SqlQueryExecution}
      * @param <R> type that should be returned by the custom sql query
-     * @return anything of type {@value R}, left to the developer, implementing the custom sql query.
+     * @return anything of type {@value <R>}, left to the developer, implementing the custom sql query.
```
pgesek: why value here, and param above?

platform/mds/mds/src/main/java/org/motechproject/mds/service/TableWriter.java
```diff
@@ -1,4 +1,4 @@
-package org.motechproject.mds.service.impl.csv.writer;
+package org.motechproject.mds.service;
```
pgesek: isn't this more of an internal class?

sebbrudzinski: The pdf/csv importer/exporter classes are created as Spring beans directly in mds-entities bundle. Both of them extend an abstract class - AbstractMdsExporter, that uses the TableWriter interface and therefore it must be exported via OSGi. I've moved it to the service package, since I made impl.csv.writer package internal (that contains implementations of TableWriter)

PR #148, opened 2015-10-16
MOTECH-1605: MOTECH startup should detect if ActiveMQ is working

atish160384:
MOTECH should display a big red alert box complaining about that on startup.We might also want a "Test ActiveMQ Connection" button in the bootstrap dialog.

PR #396, opened 2016-03-03
MOTECH-1605 MOTECH startup should detect if ActiveMQ is working

atish160384:
Motech -1605.
Added the "Verify AMQ Connection" button which checks the Active Mq connection and also displays error message when connection is not established.

### DIFF
--- a/platform/server/src/main/webapp/static/js/bootstrap-page.js
+++ b/platform/server/src/main/webapp/static/js/bootstrap-page.js
@@ -2,66 +2,132 @@ function setSuggestedValue(id, val) {
     document.getElementById(id).value = val;
 }
 
-function verifyDbConnection() {
+function setSuggestedValueByName(name, value) {
+	$("input[name='"+name+"']").val(value);
+}
+
+function saveBootstrapData(){
+    var bootstrapString = JSON.stringify($('form[name="bcform"]').serializeArray());
+    sessionStorage.setItem("bootstrapString",bootstrapString);
+};
+
+(function retrieveBootstrapData(){
+		$(document).ready(function(){
+	    		if(sessionStorage.getItem("bootstrapString") != null){
+				var bootstrapObj = JSON.parse(sessionStorage.getItem("bootstrapString"));
+				for (var key in bootstrapObj){
+				       setSuggestedValueByName(bootstrapObj[key]["name"], bootstrapObj[key]["value"]);
+				}
+				sessionStorage.removeItem("bootstrapString");
+			 }
+		});
+
+})();
+
+
+function verifyConnection(url) {
     var loader = $('#loader');
     loader.show();
 
-    var warnings = $('#verify-alert');
+    var sqlWarnings = $('#verifySql-alert');
+    var amqWarnings = $('#verifyAmq-alert');
     var infoSql = $('#verifySql-info');
-    var errors = $('#verify-error');
+    var infoAmq = $('#verifyAmq-info');
+    var sqlErrors = $('#verifySql-error');
+    var amqErrors = $('#verifyAmq-error');
+    var bootstrapErrors = $('#bootstrapErrors');
 
-    errors.html("");
-    warnings.html("");
+    sqlErrors.html("");
+    amqErrors.html("");
+    sqlWarnings.html("");
+    amqWarnings.html("");
+    bootstrapErrors.html("");
 
-    warnings.hide();
-    errors.hide();
+    sqlWarnings.hide();
+    amqWarnings.hide();
+    sqlErrors.hide();
+    amqErrors.hide();
     infoSql.hide();
+    infoAmq.hide();
+    bootstrapErrors.hide();
 
     $.ajax({
         type: 'POST',
-        url: 'verifySql',
+        url: url,
         timeout: 8000,
         data: $('form.bootstrap-config-form').serialize(),
         success: function(data) {
             if (data.success !== undefined && data.success === true) {
-                infoSql.show();
-                $(window).scrollTop(infoSql.offset().top);
+                if(url == VERIFY_SQL_URL) {
+
+                                                     infoSql.show();
+                                                     $(window).scrollTop(infoSql.offset().top);
+                                                }
+                                                if (url == VERIFY_AMQ_URL) {
+                                                    infoAmq.show();
+                                                    $(window).scrollTop(infoAmq.offset().top);
+                                                }
             } else {
-                if(data.errors !== undefined) {
+                if(data.errors !== undefined && data.sqlConfigError === true) {
                     data.errors.forEach(function(item) {
-                        errors.append(item + '<br/>');
+                        sqlErrors.append(item + '<br/>');
                     });
 
-                    errors.show();
+                    sqlErrors.show();
+                }
+                if(data.errors !== undefined && data.amqConfigError === true) {
+                    data.errors.forEach(function(item) {
+                        amqErrors.append(item + '<br/>');
+                    });
+
+                    amqErrors.show();
                 }
 
-                if(data.warnings !== undefined) {
+                if(data.warnings !== undefined && data.sqlConfigError === true) {
                     data.warnings.forEach(function(item) {
-                        warnings.append(item + '<br/>');
+                        sqlWarnings.append(item + '<br/>');
                     });
 
-                    warnings.show();
+                    sqlWarnings.show();
+                    $(window).scrollTop(warnings.offset().top);
+                }
+                if(data.warnings !== undefined && data.amqConfigError === true) {
+                    data.warnings.forEach(function(item) {
+                        amqWarnings.append(item + '<br/>');
+                    });
+
+                    amqWarnings.show();
                     $(window).scrollTop(warnings.offset().top);
                 }
             }
             loader.hide();
         },
         error: function() {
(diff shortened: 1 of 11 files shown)


frankhuster:
Needs rebase

llavoie:
@atish160384 per @frankhuster 's comment on the old PR... Does this new PR address the feedback that was left on #148?

atish160384:
I will ask srujan to comment below new PR for changes made .....

frankhuster:
Still needs rebase, still needs to incorporate some of @pgesek's comments on the older PR, and I think you want to organize the buttons in a nicer way. Maybe @nickdotreid can give an opinion?

screen shot 2016-03-08 at 1 01 57 pm

frankhuster:
Hey @atish160384, I haven't seen anything from srujan (he seems to not have a github account) about this. And since right now it's got your name on it, can you drive that? Thanks man!

srujan92:
Hi Frank,

Changes are made as suggested by Pawel in old PR. 
MotechBrokerPingUtil class is removed and its content is merged into the MotechBrokerServicePingImpl. Changes were also done in messages.properties and bootsrap-page.js as suggested.

If you have notice any other changes I might have missed, please specify.

Regarding the organizing the buttons, I have message @nickdotreid for his opinion.

P.S - I will rebase the code, by tomorrow, there's a problem with dev system which will be resolved by tomorrow.

llavoie:
Hi @srujan92 Just a heads up that @frankhuster is out of office and will be back next week. @nickdotreid just arrived back from a business trip and is probably getting himself unburied but I'm sure he'll weigh in on the button arrangement soon. In the meantime, looks like some conflicts still need to be resolved. Thanks!

nickdotreid:
@frankhuster @llavoie @atish160384 
Can I feature creep for a second: why don't we verify the connections when the form is submitted (and cancel the form submit if either fail?) — let's take human error out of the bootstrap process...

@srujan92 @atish160384 
Yes, these buttons are a little wacky.... and I think the root of the problem is more that the buttons should be close to their respective settings. A note here is to also move the error messages for the verify buttons so they are next to the verify buttons....

image

That stated, both verify buttons should be gray... the 'continue' button should be .btn-primary & .pull-right

@srujan92 I didn't get a separate message from you (at least not that I've found (minus this thread)), Where did you ping me?

llavoie:
@nickdotreid @frankhuster @srujan92 
My understanding is that the verification *does* run when you click continue -- and the bootstrap will fail if either connection doesn't pass verification. Assuming that is correct, should we get rid of the buttons to verify explicitly?

nickdotreid:
@llavoie I'm 90% sure I've started MOTECH without a database connection and then had to dump the .config files and start again — should I double check this, and if so file a bug?

llavoie:
That would be great!

-----Original Message-----

pgesek:
@llavoie @nickdotreid up to this point we didn't verify anything on submit here. I guess the idea was that they can just set the server configuration and restart later or something, not really sure. Probably makes more sense to just deny such settings as you say, unless someone has a valid use case for the current state.

nickdotreid:
@pgesek @llavoie  We really should be validating this input, allowing someone to start a broken MOTECH instance just seems silly... and I think this is a separate ticket

Let's get the verification buttons properly placed in this ticket, and make sure the callbacks work... the next ticket should validate on the backend, which I think makes more sense than validating just in javascript....

llavoie:
Yep, I agree with the approach. Does you already create the new ticket?

-----Original Message-----

nickdotreid:
I'm typing it as we speak....

atish160384:
@frankhuster @llavoie Hi, Srujan will be on leave this week. He will be coming on 28th

llavoie:
@srujan92 @nickdotreid Are we ready to merge this or are more changes needed?

nickdotreid:
@llavoie @srujan92 

The verify buttons are still placed next to each other. As I drew, they should be separated and near their original inputs... which means error messages should also get moved...

A quick/simple fix would be to make the verify buttons gray (btn-default) and the submit button blue (btn-primary) -- and the buttons should all fit in one line, or the verify buttons should be a group and pull-left while the primary submit button should pull right.

There should be another ticket down the road to simplify this page (when we split out the UI) -- and this page should not use such a large UI stack (as it currently does)

srujan92:
@llavoie @nickdotreid 

I have placed the placed the buttons near their respective settings with error message next to them, but I need to make changes to error message location when submit button is clicked. I will update the changes in the branch now.

I have made the verify buttons grey and submit button blue. But all the buttons are right.

nickdotreid:
awesome! (i was just going off the PR, so sorry if I didn't see your
changes)

srujan92:
@nickdotreid  I haven't updated the changes in atish repo just yet, I have some conflicts I need to resolve.

shanmukhm:
@nickdotreid  Please do check the changes I commited.
Thank you.

srujan92:
@nickdotreid  Please check the changes I have made.
Thank you.

platform/server/src/main/webapp/static/js/bootstrap-page.js
```diff
@@ -2,30 +2,51 @@ function setSuggestedValue(id, val) {
...
-        url: 'verifySql',
+        url: url,
         timeout: 8000,
         data: $('form.bootstrap-config-form').serialize(),
         success: function(data) {
             if (data.success !== undefined && data.success === true) {
-                infoSql.show();
-                $(window).scrollTop(infoSql.offset().top);
+                if(url == VERIFY_SQL_URL) {
+<<<<<<< HEAD
```
nickdotreid: pretty sure "<<<<<<< HEAD" is invalid javascript....

nickdotreid:
@srujan92 -- this looks about right... please see note about the merge conflict message getting committed...

srujan92:
@nickdotreid  I have cleaned the js file. 

I do have something to ask. Since we are splitting the location of error and warning messages of sql and amq. Should I split the static variable ERRORS(WARNINGS) into SQL_ERROS(SQL_WARNINGS) and AMQ_ERRORS(AMQ_WARNINGS).

nickdotreid:
@srujan92 That doesn't should like a bad idea to me... What I like is that it would clean up the JSP code a little (value string checking just looks funky)

My devil's advocate argument is that by splitting the list/collection of errors up, is that we would be making more variables for someone to track if they were going to debug the system... which I think would get uglier if we decided to add more variables to validate...

Would it make sense to split the difference? Maybe create a boolean values (sqlConfigError and amqConfigError) and pass those directly to the JSP template... that way we don't do string checking in the JSP template, and can keep fugly string checking logic in the controller with out creating many new variables...

but I totally trust your judgement... do what you think will be most maintainable for someone with no knowledge of the controller/JSP page to understand what is happening

srujan92:
@nickdotreid  I made these changes considering your suggestion. In this I have used boolean variables sqlConfigError and amqConfigError in BootstrapController. I have written the code I think could be understandable by most. Debugging won't be confusing.

This now shows errors and warnings at their correct locations unlike previous commit which sometimes show some sql ones in amq  and vice versa.

Please do review and input your suggestions.

nickdotreid:
@srujan92 
When I submit a form and ActiveMQ isn't valid, I loose all my data... which is annoying, and we should either:
(a) Return the data to the form so the user can edit/verify
(b) Verify both connections on form submit with javascript, and if both pass, then submit the form

Also, now that I'm looking at the main error response, can we move that up to the top of the form block... there is no reason it should show up below the submit button

For reference, when I first submitted the form with incorrect data, I had the window small-ish, and couldn't even see the error message (it felt like MOTECH just broke and lost my data)

image

srujan92:
@nickdotreid I have changed the main error response.
The data in Bootstrap form is also returned to their respective fields.
Please do check and comment.

nickdotreid:
@srujan92 Thanks for moving the error form location. I'm willing to call this issue patched. Could you rebase and squash these commits before I merge it?

I think there is a second UX error in this thing when the form is submitted and we loose all the form values (which should be accessible from the POST data). I don't feel that this issue is related to this ticket or urgent, so I'm going to file it as a different JIRA ticket.

srujan92:
@nickdotreid  Do you want me to remove the part where I used 'session storage' in bootstrap-page.js for retrieving form-data since it has to be done in another ticket.

nickdotreid:
@srujan92 I didn't realize that was there... it looks like all that need to happen is saveBootstrapData() needs to be called on form submit

In 'perfect-land' i'd rather not use javascript... but this is close enough

Could you: 
- call saveBootstrapData on form submit
- squash down the commit history
- get this branch in-sync with master

then I'll merge the sucker..

srujan92:
@nickdotreid  Oh sorry, somehow I forgot to add it to the submit button on the files I committed, I will do it along with squashing the commit history.

srujan92:
@nickdotreid  I have squashed changes into one commit, please do check and merge it to master.

nickdotreid:
@srujan92 -- I'm going to merge this ticket, and then file two bugs because these are small issues but don't actually reflect this ticket

After verifying (and failing) a mysql connection... a loading symbol hangs out by the submit button
image

After the page refreshes, all the inputs need to be touched once to become what ngForm thinks of as $dirty (and then lets the form be submitted)... this is an odd edge case....

PR #398, opened 2016-03-04
MOTECH-747 MOTECH Data Services (MDS) architecture topic

atish160384:


docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,109 @@
-Data Services Architecture
-==========================
+Motech Data Services Architecture
+=================================
 
-There will be text here.
+MOTECH Data Services (MDS) is responsible for creation of Data Models for MOTECH Platform either using Developer Defined Entities (DDE) or End User Defined Entities (EUDE). DDEs use custom annotations of MDS. The EUDEs can be created as well as configured from UI along with their Data Access Layer. Centre of Data Access Layer is Generic Data Access Pattern Class which defines generic persistence related methods.
```
llavoie: "DDEs use custom annotations of MDS" is not very clear. Maybe something like "DDEs use custom annotations of Java objects to specify entities and fields"...

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,109 @@
...
+            :align: center
+
+
+        Section following describes motech data services architecture in figure.
+
+MDS UI Layer
+============
+
+#.Schema Editor: It is responsible for browsing through entities.
+#.Data Browser: It is responsible for browsing through entities.
```
llavoie: From these descriptions, one would think that the Schema Editor and the Data Browser serve the same purpose. Can you elaborate on what they are for?

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,109 @@
...
+#.Settings: It is responsible for setting data retention policies .i.e. managing trash and history data.It also provides an option to export and import schema using csv as Input or output.
```
llavoie: There are some problems with dots and spacing in these couple of sentences. There should be no space between the end of a sentence and its period. There should be one space between a period and a new sentence.

llavoie: Inconsistent capitalization "Input" and "output" -- I would prefer that they are both lowercase.

atish160384:
@llavoie I have made changes in file as told accordingly. Please do check

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
-Data Services Architecture
-==========================
+Motech Data Services Architecture
+=================================
 
-There will be text here.
+MOTECH Data Services (MDS) is responsible for creation of Data Models for MOTECH Platform either using Developer Defined Entities (DDE) or End User Defined Entities (EUDE). DDEs use custom annotated POJO mark them as MDS entities. The EUDEs can be created as well as configured from UI along with their Data Access Layer. Centre of Data Access Layer is Generic Data Access Pattern Class which defines generic persistence related methods.
```
pgesek: mark them -> to mark them

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+
+        Section following describes motech data services architecture in figure.
+
+MDS UI Layer
+============
+
+#.Schema Editor: It is responsible for creating new EUDEs and also to edit already existing entities (Adding new fields and their lookups).
+#.Data Browser: It is responsible for browsing through entities.
+#.Settings: It is responsible for setting data retention policies .i.e. managing trash and history data. It also provides an option to export and import schema using csv as input or output.
```
pgesek: 1) It allows exporting schema or data
2) The format is json, not csv

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+MDS Web
+=======
+
+It acts as controller between MDS UI Layer and MDS Core Layer to view, create and update EUDE and it’s infrastructure classes .i.e services and repositories and to view and update DDE and it’s infrastructure classes .i.e services and repositories. It use pre defined services of MDS.
```
pgesek: a controller

pgesek: It use -> It uses

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+MDS Core Layer 
+==============
+
+#. MDS Initializer: It is responsible for building all entites in MDS database. It use MDS Constructor for building entites. It’s using TransactionalCallBackWithout Result of Spring framework for configuring Transaction. It includes MDSWeavingHook which is responsible for hooking into OSGi class loading process to dynamically add Jdo Imports and replacing DDE’s with thier bytecodes (extended version) generated by MDS.
```
pgesek: It's actually more responsible for starting the core classes of MDS such as the MDS bundle watcher. The initializer does no longer directly interact with the constructor (the javadoc is outdated). Also worth noting is that the initializer posts an OSGi event after MDS initialization is complete - the event is used to control the startup process of the entire platform.

pgesek: Moreover, 'includes the weaving hook' is not entirely correct, it only registers it in OSGi upon startup, the weaving hook is a separate class.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+Added Jdo Imports are as follows:-
```
pgesek: Don't think we need to explicitly specify all added imports, this is a technical detail, that can keep changing. Not important to the architecture overall

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+javax.jdo,javax.jdo.identity,javax.jdo.spi,org.joda.time,org.apache.commons.lang,org.springframework.transaction.support,org.motechproject.mds.filter,org.motechproject.mds.query,org.motechproject.mds.util,org.motechproject.commons.date.util,org.datanucleus.enhancer,org.datanucleus,org.datanucleus.state,org.datanucleus.asm,org.datanucleus.exceptions,org.datanucleus.metadata,org.datanucleus.plugin,org.datanucleus.util,org.datanucleus.identity
+
+
+#. MDS Constructor: It provides methods to create entities, their service implementations and corresponding interfaces.
+
+#. Entity Builder: It is use to build classes of a given entities.
```
pgesek: of all entities in MDS. It uses javassits for manipulating bytecode

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+MDS Core Layer 
+==============
+
+#. MDS Initializer: It is responsible for building all entites in MDS database. It use MDS Constructor for building entites. It’s using TransactionalCallBackWithout Result of Spring framework for configuring Transaction. It includes MDSWeavingHook which is responsible for hooking into OSGi class loading process to dynamically add Jdo Imports and replacing DDE’s with thier bytecodes (extended version) generated by MDS.
+
+Added Jdo Imports are as follows:-
+javax.jdo,javax.jdo.identity,javax.jdo.spi,org.joda.time,org.apache.commons.lang,org.springframework.transaction.support,org.motechproject.mds.filter,org.motechproject.mds.query,org.motechproject.mds.util,org.motechproject.commons.date.util,org.datanucleus.enhancer,org.datanucleus,org.datanucleus.state,org.datanucleus.asm,org.datanucleus.exceptions,org.datanucleus.metadata,org.datanucleus.plugin,org.datanucleus.util,org.datanucleus.identity
+
+
+#. MDS Constructor: It provides methods to create entities, their service implementations and corresponding interfaces.
```
pgesek: The constructor calls entity builder and the metadata builder, it also does some stuff with the Datanucleus Enhancer. We should expand this part.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Entity Builder: It is use to build classes of a given entities.
+
+#. Infrastructure Builder: It is responsible for building infrastructure for a given entity .i.e repository, interface and service classes. These classes are created only if not available in the classpath using  javassist to construct them.
```
pgesek: These classes are always built, the only exception might for service interfaces already defined in bundles. Classpath doesn't have much to do with it.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Metadata Builder: It is responsible for generating JDO metadata for entity.
```
pgesek: I think we can expand a bit more on this - for example that it uses both the schema and the class definition in case of the DDE for coming up with the result schema.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Enum Builder: Value defined in Combo Box/ Field Metadata will be used to create an Enum.
```
pgesek: But only if it does not allow users to provider their own data (in which case it will be a String)

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Lookup Builder: It allows to create a method without body and added into javassist service interface and method with body added into javassist service class.
```
pgesek: this 'method without body' is just a method declaration in the interface. Can we expand a bit more on what these methods do and what is their correlations with lookups? For example each of these methods represents a lookup, and the implementation code which is generated for it calls a method from the base class. Each lookup parameter is a parameter of the lookup method, the return type is dependent on whether the lookup returns multiple values and so on

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Processors: Processors process DDEs based on annotations and create Infrastructure and Metadata Information for those entities which can used via DataNucleus enhancer to enhance entities to use them for Persistence Operations.
```
pgesek: can used -> can be used

pgesek: Mention that processors use reflections to scan bundles after they are installed (MdsBundleWatcher keeps track of bundles getting installed). The information that is gathered by processors goes to the schema database, which along EUDE information is used by the JarGeneratorService (which delegates much of this to the MDSConstructor) to build the MDS classes, metadata and infrastructure

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. JDO Enhancer: It extends DataNuclues JDOEnhancer. It enhances classes to add missing information so that the classes can be used by DataNuclues for persistance related operations.
+
+#. MDS Data Provider: It creates Mds data provider json to register in task module.
+
+#. Infrastructure Classes, EUDE and DDE: It represents all entities, services, repositories of entities used by MDS Web module to represent MDS entities in UI.
+
+#. Pre Defined Service Section:
```
pgesek: This should fit more along the data provider service

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Lookup Builder: It allows to create a method without body and added into javassist service interface and method with body added into javassist service class.
+
+#. Processors: Processors process DDEs based on annotations and create Infrastructure and Metadata Information for those entities which can used via DataNucleus enhancer to enhance entities to use them for Persistence Operations.
+
+#. JDO Enhancer: It extends DataNuclues JDOEnhancer. It enhances classes to add missing information so that the classes can be used by DataNuclues for persistance related operations.
+
+#. MDS Data Provider: It creates Mds data provider json to register in task module.
+
+#. Infrastructure Classes, EUDE and DDE: It represents all entities, services, repositories of entities used by MDS Web module to represent MDS entities in UI.
```
pgesek: These classes live in their own generated bundle, and mentioning mds web here might be misleading. The service interfaces and entity classes are used by all modules, not only mds-web, to deal with the persistence layer. Repository classes are internal and not available outside the generated bundle.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+#. Pre Defined Service Section:
+Mds provide some predefined services used by Mds Web and other modules of motech platform :
+
+    *Action Handler Service: It provides methods for handling tasks actions events related
+    with MDS CRUD operations .i.e. create, update and delete.
+
+    * Bundle watcher Suspension Service: An OSGi service, allowing to temporarily disable bundle processing by MDS. When processing gets suspended, MDS will still listen to bundle events, but instead of processing them, they will be queued and processed after the processing gets restored. This allows, besides others, to install/uninstall a larger amount of bundles at one time, without facing annotation processing problems.
```
pgesek: This is a very important component that does a lot of stuff under the covers, such as calling the annotation processor, using the JarGeneratorService to generated the bundle and so on. It should have its own section,

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Csv Export Customizer: It provides custom methods to format related instances during csv export.
```
pgesek: The customizers are not services, just parts of the API contract

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Csv Import Customizer: It provides custom methods for searching, creating and updating instances during csv import.
+
+    * Csv Import Export Customizer: Service providing methods for importing and exporting instances using csv.
+
+    * Motech Data Service: Centre of Data Access Layer is Generic Data Access Pattern Class which defines generic persistence related methods. Superclass for defining service of any Entity in MDS.
```
pgesek: I think that the way services are constructed, what they inherit and what do they use should be a bigger part of this doc. This is too little.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Entity Service: It defines CRUD actions on entities.
```
pgesek: It allows manipulating the entity schema and it back the schema editor UI. It is import to mention that this is for the schema, not instances.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * History Service: It persists changes made on given instance. It captures instances which are not in trash.
```
pgesek: I don't get the trash part. It's also worth noting that the history service and trash service are called from JDO lifecycle listeners to persist trash/history instances on delete/update.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Trash Service: It provides methods related to trash mode (By Default, trash mode is active and can be turned off by user.)
+
+    * Jar Generator Service: It creates a bundle jar with all entities defined in MDS.
```
pgesek: This also should be a prominent point in the doc - for generating the bundle this service for example uses Velocity templates to generate proper Spring XML files for the entities bundle.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Json LookUp Service: Service for managing lookups from JSON.
```
pgesek: What are lookups from json?

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * MDS Bundle Regeneration Service: It provides methods to regenerate bundle and regenerate bundle after dde enhancement. This service use Jar Generator Service to regenerate bundle.
```
pgesek: On the topic of this service, we should describe how we use an event topic to make sure all clustered MDS instances regenerate their schema - which is the only real purpose of this service.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * MDS LookUp Service: It provides a methods for processing lookups for particular entity given names, class names as well as String names. It provides a generic access to entities.
+
+    * MDS Scheduler Service: It provides methods to schedule and unschedule a repeating job. It used for emptying trash periodically.
```
pgesek: This is pretty much internal to MDS and the reason for using quartz directly in MDS is to avoid circular dependency issues (mention that).

pgesek: It used -> It is used

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Migration Service: It provides method to find flyway migrations within bundle (Default Search location is db/migration).
```
pgesek: bundle -> bundles

pgesek: Describe how flyway migrations in modules (and flyway migrations in general) are handled by MDS

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * Type Service: It provides methods for defining methods for executing actions on type.
+    Rest Documentation Service. It provides functionality of dynamically generated MDS Rest Documentation.
```
pgesek: Bad formatting with the Rest Documentation service

pgesek: It generates a swagger json file based on the entity schema

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+    * MDS Bundle Regeneration Service: It provides methods to regenerate bundle and regenerate bundle after dde enhancement. This service use Jar Generator Service to regenerate bundle.
+
+    * MDS LookUp Service: It provides a methods for processing lookups for particular entity given names, class names as well as String names. It provides a generic access to entities.
+
+    * MDS Scheduler Service: It provides methods to schedule and unschedule a repeating job. It used for emptying trash periodically.
+
+    * Migration Service: It provides method to find flyway migrations within bundle (Default Search location is db/migration).
+
+    * Type Service: It provides methods for defining methods for executing actions on type.
```
pgesek: From my understanding type service simply provides access to the types defined in MDS.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+    Rest Documentation Service. It provides functionality of dynamically generated MDS Rest Documentation.
+
+    * JDO Listener Registry Service: It provides access to registry of Listeners for Persistence events .i.e. POST_CREATE, PRE_DELETE, POST_DELETE, POST_LOAD, PRE_STORE and POST_STORE.It handles pre and post events of crud events of DataNuclues.
```
pgesek: We should describe what this registry does and how (it picks up properly annotated OSGi services and hooks them to JDO lifecycle events)

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+
+
+
+Notes
+-----
+*  Custom Annotations, their processing and actions performed on those annotations is a part of MDS. These processors process DDEs and create Metadata as well as Infrastructure Information for those entities which can used via DataNucleus enhancer to enhance entities to use them for Persistence Operations.
```
pgesek: wasn't this already mentioned in the processors section?

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+*  EUDEs, the related Data Access Layer and their related MetaData Information is generated which can be used by DataNucleus for Persistence Operations.
+
+*  Above mentioned Entities and related Data Access Layer and Metadata information are created during bootstrapping of MOTECH Platform.
+
+*  Whenever an EUDE is defined from UI, 'MOTECH Data Services' bundle is regenerated.
```
pgesek: 1) Not 'MOTECH Data Services' (which is mds) is regenerated and restarted - it's MOTECH Platform Data Services Entities
2) This doc lacks information about the generated bundle (so does the diagram), while it is a very important of the architecture. Even a very high level look at the MDS architecture must include this - all the generated classes and infrastructure go into this bundle. We should describe how this bundle is managed and how OSGi is used here.

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+*  Whenever EUDE lookups are defined using UI, 'MOTECH Data Service' bundles is regenerated.
+
+* MDS also provides features to create REST APIs for creating, updating, deleting and searching the instances of the Entities created.
```
pgesek: Needs more information on how this works (rest facades etc.)

docs/source/architecture/data_services_architecture.rst
```diff
@@ -1,4 +1,110 @@
...
+
+* Feature To download/upload schema using csv as input or output.
```
pgesek: This is out of place

pgesek: This doc seems to lack information about the databases being used - MDS uses two databases now.

llavoie:
@atish160384 Will you or Srujan be fixing this up? (I don't know Srujan's github handle)

srujan92:
@llavoie  I will fix this.

PR #399, opened 2016-03-04
MOTECH-1163: Merge platform and module directories in platform-modules

lkukielka:


pgesek:
This is outdated. Moreover I don't think we should care about this ticket for now - it will do more chaos than it is worth

PR #400, opened 2016-03-04
MOTECH-2244 Errors removed from console

lkukielka:


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -1549,6 +1549,9 @@
         /* ~~~~~ FIELD FUNCTIONS ~~~~~ */
 
         $scope.isUniqueEditable = function(field) {
+            if (field.type === undefined) {
```
pgesek: just do if (field.type) same for other ifs

PR #401, opened 2016-03-04
MOTECH-2063: Improved datetimepicker

ngraczewski:


modules/scheduler/scheduler/src/main/resources/webapp/js/controllers.js
```diff
@@ -209,6 +209,14 @@
             return date;
         }
 
+        $scope.getMinDate = function(jobType) {
```
sebbrudzinski: Why do we only enforce starting date for run once jobs? Is this for convenience when updating jobs?

ngraczewski: The reason for that is that trying to schedule a job with a start date in the pass causes server side error(there is no point in scheduling job that will never be executed). Other job types can have their start date set in the past, because all of them will be repeated multiple times. User can chose to ignore the executions that would happen before current date with "Ignore past fires at start" property at the job creation screen.

PR #402, opened 2016-03-04
MOTECH-1747 - DEV - BUG - CSV Import feature - We are not controlling…

atish160384:
… the lenght of the name

Added a css class with eplipsis to handle this issue.


kmadej:
@atish160384 this problem regards import CSV feature. You try to add new css to import.html file, which is used in import schema in MDS. 
screenshot from 2016-03-09 13 52 49

You are supposed to make changes in import-entity.html file. This file is used in import CSV feature, which you can find in Data Browser -> choose an entity -> 'Import CSV' button

Please next time check carefully if your changes fix the problem described on JIRA.

atish160384:
I have added the css class to import-entity.html and removed it from import.html

kmadej:
It works as expected (checked on firefox and chrome) :
screenshot from 2016-03-10 13 58 02

PR #403, opened 2016-03-04
MOTECH-2094: clean up scheduler API

sebastianceronik:
1. Deleted mds.service and mds.service.impl imports wich were not used in scheduler
2. some code and javadocs cleanup to keep consistency within class/package
3. changed concatenation to use stringBuilder rather than "+" operator
4. deleted deprecated function from API (updateScheduledJob) it was deprecated for way more than a year.
5. removed author from class javadocs 
6. made few changes in event module since I overlooked them while doing "MOTECH-2088"

I wasn't sure if we need to export org.motechproject.scheduler.builder
it might get in handy export this package to create jobCronExpressions


sebastianceronik:
removed unused import

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/CronJobSimpleExpressionBuilder.java
```diff
@@ -28,6 +28,7 @@ public CronJobSimpleExpressionBuilder(Time startTime) {
      * Sets interval on which job should be fired.
      *
      * @param repeatIntervalInDays  the interval(in days) between job fires, 0 means everyday
+     *
```
kmadej: new line between description and param is ok, however param, return and throws should be without new line. Follow the javadoc conventation used in mds module. 

This comment regards other changes

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerService.java
```diff
@@ -45,15 +40,15 @@
...
-     * @param job  the job DTO
+     * @param job the SchedulableJob
      */
     void scheduleJob(SchedulableJob job);
 
     /**
-     * Schedules the given schedulable job. The Job ID by which the job will be referencing in the future should be provided
-     * in an Instance of MotechEvent in SchedulableJob (see MotechEvent.jobId)
+     * Schedules the given cron schedulable job. The Job ID by which the job will be referencing in the future should be provided
+     * in an Instance of MotechEvent in SchedulableJob (see motechEvent.getParameters().get("JobID"))
```
kmadej: the tag 'see' should indicate other method/variable. After changes in scheduler I think this see should be removed

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/MotechSchedulerService.java
```diff
@@ -1,7 +1,6 @@
 package org.motechproject.scheduler.service;
```
kmadej: add missing description to param in javadoc, for example in method 'safeUnscheduleRunOnceJob'. Every param should have brief description.

kmadej:
Remember about javadoc for public methods in every exported class. I noticed that in package contract, few classes and methods don't have javadoc.

kmadej:
PR closed because of open new one. More information in https://github.com/motech/motech/pull/413

PR #404, opened 2016-03-05
Added generated test bundle to .gitignore

sebbrudzinski:
For the bundle upload integration test we
now generate a test-bundle, directly in the
admin test resources. This auto-generated file
should not get staged for commits, thus adding
it to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ tools/SMPPSim/
 *.tmp
 platform/server/src/main/webapp/WEB-INF/classes/
 platform/server/src/main/webapp/[0-9]*/
+modules/admin/src/test/resources/motech-upload-test-bundle.jar
 docs/build
 docs/env
 docs/source/packages.rst


pgesek:
Why are we placing this in resources anyway? Why not use the project build directory?

sebbrudzinski:
@pgesek The clean step would have to be reconfigured in the admin module to not throw away the whole target directory, which it currently does. I don't think it would be a huge problem to change that, but is there any reason we wouldn't want to use resources directory?

pgesek:
@sebbrudzinski 

Well it's not great practice to be putting generated files into the resources directory, which is a source directory. It's the same reason why we are not putting .class files in the src/java directory.

Not entirely sure what the clean step has to do with this, doesn't it run at the very beginning?

sebbrudzinski:
@pgesek 
It runs at the very beginning of module installation (admin in our case). The bundle is generated as a step of another module's build though (upload-test-bundle here), which must run before admin.

pgesek:
In that case this is completely backward - we don't want to have such hidden dependencies between modules. Basically, if I run mvn clean install -PIT in a module it should work, I shouldn't be required to build some hidden dependency or copy a file beforehand. We should use the dependency plugin to copy the test bundle dependency to the target directory, not expect that it will be in the resources dir.

PR #406, opened 2016-03-06
MOTECH-2124 Modifed Authentication Handler for JSON Login

nickdotreid:
> Added logic to authentication handlers that catches XMLHttpRequests, and
> returns text, instead of redirecting. This was created to support a
> seperated UI that would communicate with MOTECH via JSON.

There is still a little bit of work I need to include, such as
- Returning a user object on successful authentication
- Returning an error message depending on why the login didn't work (ie User Blocked)
- Adding tests...

### DIFF
--- a/platform/web-security/src/main/java/org/motechproject/security/authentication/MotechLoginErrorJSONHandler.java
+++ b/platform/web-security/src/main/java/org/motechproject/security/authentication/MotechLoginErrorJSONHandler.java
@@ -1,0 +1,26 @@
+package org.motechproject.security.authentication;
+
+
+import org.motechproject.osgi.web.extension.HttpRequestEnvironment;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.authentication.ExceptionMappingAuthenticationFailureHandler;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class MotechLoginErrorJSONHandler extends ExceptionMappingAuthenticationFailureHandler {
+
+    @Override
+    public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response, AuthenticationException exception)
+            throws IOException, ServletException {
+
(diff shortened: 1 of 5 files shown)


nickdotreid:
@pgesek -- can you let me know if I'm putting things in the right place?

platform/web-security/src/main/java/org/motechproject/security/authentication/MotechLoginErrorJSONHandler.java
```diff
@@ -0,0 +1,26 @@
...
+import org.motechproject.osgi.web.extension.HttpRequestEnvironment;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.authentication.ExceptionMappingAuthenticationFailureHandler;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class MotechLoginErrorJSONHandler extends ExceptionMappingAuthenticationFailureHandler {
```
pgesek: javadoc

platform/web-security/src/main/java/org/motechproject/security/authentication/MotechLoginSuccessJSONHandler.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.web.authentication.SavedRequestAwareAuthenticationSuccessHandler;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class MotechLoginSuccessJSONHandler extends SavedRequestAwareAuthenticationSuccessHandler {
```
pgesek: javadoc

pgesek:
@nickdotreid seems fine, just add the missing javadocs

nickdotreid:
I just realized that a more up to date PR exists at #472 

(ooops)

PR #412, opened 2016-03-08
MOTECH-2063: Separated DayOfWeek job

ngraczewski:
Separated DayOfWeek job and fixed some issues with job edition

### DIFF
--- a/modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/SchedulableJobBuilder.java
+++ b/modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/SchedulableJobBuilder.java
@@ -1,8 +1,12 @@
 package org.motechproject.scheduler.builder;
 
 import org.joda.time.DateTime;
+import org.motechproject.commons.date.model.DayOfWeek;
+import org.motechproject.commons.date.model.Time;
 import org.motechproject.event.MotechEvent;
 import org.motechproject.scheduler.contract.CronSchedulableJob;
+import org.motechproject.scheduler.contract.DayOfWeekJobId;
+import org.motechproject.scheduler.contract.DayOfWeekSchedulableJob;
 import org.motechproject.scheduler.contract.RepeatingJobId;
 import org.motechproject.scheduler.contract.RepeatingPeriodJobId;
 import org.motechproject.scheduler.contract.RepeatingPeriodSchedulableJob;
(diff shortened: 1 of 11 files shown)


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/contract/DayOfWeekSchedulableJob.java
```diff
@@ -69,19 +72,43 @@ public DayOfWeekSchedulableJob(MotechEvent motechEvent, DateTime startDate, Date
         this.time = time;
     }
 
+    /**
+     * Reruns the time of the job execution.
```
sebbrudzinski: Returns

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/SchedulableJobBuilder.java
```diff
@@ -101,7 +128,21 @@ private static SchedulableJob buildRunOnceSchedulableJob() {
...
+        return new Time(Integer.valueOf(parts[2]), Integer.valueOf(parts[1]));
+    }
+
+    private static List<DayOfWeek> getDays(String cron) {
+        String[] days = cron.split(" ")[5].split(",");
+        List<DayOfWeek> daysList = new ArrayList<>();
+        for (String day : days) {
+            daysList.add(DayOfWeek.getDayOfWeek(Integer.valueOf(day)));
+        }
+        return daysList;
```
sebbrudzinski: For the above methods, can we use some simple library/cron util to do the parsing for us?

ngraczewski: I couldn't find any library/util to achieve this. I'll create separate utility class for this.

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/contract/DayOfWeekJobId.java
```diff
@@ -0,0 +1,30 @@
+package org.motechproject.scheduler.contract;
+
+import org.motechproject.event.MotechEvent;
+
+/**
+ * Jobs ID for {@link DayOfWeekSchedulableJob}s.
+ */
+public class DayOfWeekJobId extends JobId {
```
sebbrudzinski: By the way, this may be a good time to start adding suffix to cron expressions as well. Isn't it "null" for cron expressions now?

ngraczewski: No, cron job doesn't have any suffix at the moment.

PR #414, opened 2016-03-08
Fix broken parsing of Motech dir

pgesek:
Currently it creates the Motech dir as literally ${

This currently reproducible for module upload


pgesek:
test this please

PR #417, opened 2016-03-09
MOTECH-1450 Fixed asserts in AdminBundleIT

lkukielka:


sebbrudzinski:
@lkukielka please move the fixes to the new test class to resolve conflicts. Also, since this is no longer run as IT, we can go back to installing cms-lite from repository, rather than tasks (there's currently a small problem with tasks throwing exception, even though the test works properly)

sebbrudzinski:
613 files edited and conflicts. Something is not quite right here. Since those are just 2 lines to modify, perhaps it will be easier to just open a new pull request from the updated master?

lkukielka:
@sebbrudzinski Fixed

PR #418, opened 2016-03-09
Move AdminBundleIT to a separate profile

pgesek:
The IT was turned into an FT that will run
only in the FT profile. For now this is acceptable.

### DIFF
--- a/modules/admin/pom.xml
+++ b/modules/admin/pom.xml
@@ -193,10 +193,6 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-            </plugin>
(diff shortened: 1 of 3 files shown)


pgesek:
http://ci.motechproject.org/job/Motech-Admin-FT/ will build these tests daily

modules/admin/pom.xml
```diff
@@ -363,4 +359,46 @@
         </plugins>
     </build>
 
+    <profiles>
+        <profile>
+            <id>FT</id>
```
sebbrudzinski: This profile should also handle the copying and cleaning of the configuration on its own.

sebbrudzinski:
test this please

sebbrudzinski:
@pgesek OK to merge

PR #420, opened 2016-03-10
Added separate methods for unscheduling DayOfWeek jobs

ngraczewski:


sebbrudzinski:
It seems like the path we have chosen is dead end and introduces new problems in more places. Let's revert the changes to day of week and cron suffixes and simply add a flag in job data to recognize the jobs for edit purposes.

PR #419, opened 2016-03-09
Added suffix for the Cron job in the Scheduler module

ngraczewski:

PR #421, opened 2016-03-10
Revert "Added suffix for the Cron job in the Scheduler module"

ngraczewski:
Reverts motech/motech#419


sebbrudzinski:
Can you do the reverts to the cron and day of weeks suffixes locally and upload one pull request that does it?

PR #422, opened 2016-03-11
Changed the way of distinguishing the DayOfWeek jobs

ngraczewski:


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/builder/SchedulableJobBuilder.java
```diff
@@ -130,14 +130,14 @@ private static DateTime getEndDate(Trigger trigger) {
...
-    private static String getJobType(JobKey jobKey) throws SchedulerException {
+    private static String getJobType(JobKey jobKey, JobDataMap dataMap) throws SchedulerException {
         if (jobKey.getName().endsWith(RunOnceJobId.SUFFIX_RUNONCEJOBID)) {
             return RUN_ONCE;
         } else if (jobKey.getName().endsWith(RepeatingJobId.SUFFIX_REPEATJOBID)) {
             return REPEATING;
         } else if (jobKey.getName().endsWith(RepeatingPeriodJobId.SUFFIX_REPEATPERIODJOBID)) {
             return REPEATING_PERIOD;
-        } else if (jobKey.getName().endsWith(DayOfWeekJobId.SUFFIX_DAY_OF_WEEK_JOB_ID)) {
+        } else if (dataMap.getBoolean(IS_DAY_OF_WEEK)) {
```
sebbrudzinski: Is this null safe? If there's no flag in the dataMap, will this evaluate as false?

ngraczewski: Yes, it is.

PR #425, opened 2016-03-11
MOTECH-2063 Fixed schedule job date validation

lkukielka:


modules/scheduler/scheduler/src/main/resources/webapp/js/controllers.js
```diff
@@ -240,6 +240,14 @@
                 job.endDate = $scope.parseToDateTime(job.endDate);
             }
 
+            if (job.startDate && job.endDate) {
+                if (job.startDate > job.endDate) {
+                    motechAlert("scheduler.error.incorrectJobDate", "scheduler.error");
+                    unblockUI();
```
ngraczewski: UI shouldn't be blocked at this point so the call of the unblockUI function is not needed.

sebbrudzinski:
@pgesek CR completed

PR #426, opened 2016-03-11
MOTECH-2063: Fixed some issues with job creation/edition

ngraczewski:


sebbrudzinski:
Please resolve conflicts

sebbrudzinski:
Pleasy verify why editing repeat jobs breaks the motechquartz database from time to time.

sebbrudzinski:
Editing repeating jobs still moves the end date forward by 1 hour. Please look into it.

sebbrudzinski:
Create a job type of any type but "Run Once". Attempt to update it, by changing end date. Save. You get "undefined" error and job is not updated.

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -651,6 +643,14 @@ private void scheduleRepeatingJob(RepeatingSchedulableJob job, boolean update) {
         jobDetail.getJobDataMap().put(IGNORE_PAST_FIRES_AT_START, job.isIgnorePastFiresAtStart());
         jobDetail.getJobDataMap().put(USE_ORIGINAL_FIRE_TIME_AFTER_MISFIRE, job.isUseOriginalFireTimeAfterMisfire());
 
+        try {
+            if (scheduler.getTrigger(triggerKey(jobId.value(), JOB_GROUP_NAME)) != null) {
+                unscheduleJob(jobId);
+            }
+        } catch (SchedulerException e) {
+            throw new MotechSchedulerException(format("Schedule or reschedule the job: %s.\n%s", jobId, e.getMessage()), e);
```
sebbrudzinski: It seems this error message is not complete. Moreover, is this message we show on UI? If so, it should be localized like the others.

PR #431, opened 2016-03-14
MOTECH-2264 Fixed restarting a module caused a null ptr excecption

lkukielka:


platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -308,7 +308,7 @@ public void addOrUpdateProperties(String bundle, String version, String filename
             //Persist only non-default properties in database
             toPersist = new Properties();
             for (Map.Entry<Object, Object> entry : newProperties.entrySet()) {
-                if (!defaultProperties.containsKey(entry.getKey()) ||
+                if (defaultProperties == null || !defaultProperties.containsKey(entry.getKey()) ||
```
pgesek: When would this be null, just curious?

sebbrudzinski: @lkukielka Can you comment on that?

lkukielka: @pgesek  Sorry for such a long delay but I didn't notice that you commented.
_defaultProperties_ would be null when user is submitting changes in non defaults properties file, i.e. sms-templates.json or atom-client-feeds.json

pgesek: why would this method get called for json files?  Isn't this only for property files?

lkukielka:
@pgesek It's not a bug. Ultimately, _addOrUpdateProperties_ is using to add or update *.properties files and also is using to update raw data files. There is also a _saveRawConfig_ method, but it is used only for adding new raw file, so I think this solution is correct. Do you think is we need to create separate method to updating raw data files?

pgesek:
I just don't get why this method would get called for json files, when it is called addOrUpdateProperties and is working on Properties objects. Who is calling it to save a json file? And how is he passing a json file to this method?

lkukielka:
Submitting both, raw data and properties, is triggering _saveSettings_ method with argument _mSettings_ which contains file name and list properties.

pgesek:
@lkukielka submitting where? which saveSettings method? Give me a precise description of the code path that is saving json files using the addOrUpdateProperties() method.

lkukielka:
First of all click _Submit_ under _rawData_


What causes the following steps:

in file motech/modules/admin/src/main/resources/webapp/js/controllers.js:643

arguments values: mSettings = $ {section: "atom-client-feeds.json", settings: Array[1], $$hashKey: "02C", $resolved: true}, doRestart = undefined

        $scope.saveSettings = function (mSettings, doRestart) {
            // 
            //unimportant lines I will ignore
            //	
            mSettings.$save({bundleId:$scope.module.bundleId}, successHandler, angularHandler('admin.error', 'admin.settings.error'));
        };

**next:**
in file motech/modules/admin/src/main/java/org/motechproject/admin/web/controller/SettingsController.java:69

arguments: bundleId = 142, bundleSettings = 


    public void saveBundleSettings(@PathVariable long bundleId, @RequestBody Settings bundleSettings) throws IOException {
        settingsService.saveBundleSettings(bundleSettings, bundleId);
    }

**next:**

in file motech/modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/SettingsServiceImpl.java:134

arguments: settings = the same as on the screen above, bundleId = 142

    @Override
    public void saveBundleSettings(Settings settings, long bundleId) throws IOException {
        Properties props = ParamParser.constructProperties(settings);

        configurationService.addOrUpdateProperties(getBundleSymbolicName(bundleId), getVersion(bundleId), settings.getSection(),
                props, getBundleDefaultProperties(bundleId).get(settings.getSection()));
    }

**next:**

in motech/platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java:305

arguments: bundle = "org.motechproject.atom-client", version = "1.0.0.SNAPSHOT", filename = "atom-client-feeds.json" newProperties = 
 defaultProperties = null

    public void addOrUpdateProperties(String bundle, String version, String filename, Properties newProperties, Properties defaultProperties) throws IOException {
        Properties toPersist;
        if (ConfigSource.UI.equals(configSource)) {
            //Persist only non-default properties in database
            toPersist = new Properties();
            for (Map.Entry<Object, Object> entry : newProperties.entrySet()) {
                if (defaultProperties == null || !defaultProperties.containsKey(entry.getKey()) ||
                        (!defaultProperties.get(entry.getKey()).equals(newProperties.get(entry.getKey())))) {
                    toPersist.put(entry.getKey(), entry.getValue());
                }
            }
        } else {
            toPersist = newProperties;
            checkDifferencesAndSaveFile(bundle, filename, toPersist);
        }
        ModulePropertiesRecord properties = new ModulePropertiesRecord(toPersist, bundle, version, filename, false);
        if (bundlePropertiesService != null) {
            addOrUpdateBundleRecord(properties);
        }

    }

**next:**

in file motech/platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java:618

argument: record =


    public void addOrUpdateBundleRecord(ModulePropertiesRecord record) {
        ModulePropertiesRecord rec = getBundlePropertiesRecord(record.getBundle(), record.getFilename());
        if (rec == null) {
            bundlePropertiesService.create(record);
        } else {
            rec.setProperties(record.getProperties());
            bundlePropertiesService.update(rec);
        }
    }

Above BundlePropertiesService extends MotechDataService and finally update method save json.

pgesek:
Ok I see. I don't like that we are storing these raw files as properties underneath, but refactoring all of this is probably a separate ticket for another day.

pgesek:
This needs conflicts to be resolved

lkukielka:
Ok, finally I removed conflicts, sorry for delay

PR #437, opened 2016-03-17
MOTECH-2297: Prevents double listener registration for proxied beans

kmadej:
# **## Overview**
In 'problem description' I describe briefly how it works and how my changes fix the problem. If someone is more curious in section 'understanding proxy usage in Spring' I describe more precisely how it works (pictures and information are from https://spring.io/blog/2012/05/23/transactions-caching-and-aop-understanding-proxy-usage-in-spring). Example 3 and 4, which are mentioned in briefly description are in section 'Understanding proxy usage in Spring'.

## **## Problem description**
If someone annotates method with @MotechListener, the event module will register this method as a listener for subject. By default there will be normal object under the spring bean. However if in the bean is at least one @Transactional annotation (there is no need this two annotations are on the same method), under the bean will be spring proxy (only one even if service is exposed as an OSGi service). Then event module will register all methods annotated with @MotechListener from this proxy and from its superclasses. 

Problem occurs when our bean is constructed similar to example 4. Then two listeners will be registered. All proxy methods are marked as final, so we will have one method from proxy(final), second from proxy parent(not final), which is our real object.

Problem doesn't occur in example 3. Because proxy implements (not extends) our bean. So we will have only one listener - method marked as final in proxy.   

I resolved this double listener registration. Now the event module seeks listeners only in object methods (without superclass methods).
## 
## **## Understanding proxy usage in Spring**
In the Spring framework, some features rely on proxy usage (for example transactions). I will try to explain here how exactly it works.

**Example 1** - Let's assume we've got simple spring component :
```
@Component
public class TestComponent {
}
```

under spring bean there is no proxy, just simple calling to object of type TestComponent

**Example 2 **- Spring service 
```
@Service
public class AccountServiceImpl implements AccountService {
}
```

At runtime :

no-proxy

In MOTECH even if a service is exposed as an OSGi service there will be no proxy under spring bean.

**Example 3** - Situation is a bit different when we add the spring @Transactional annotation to bean :
 
```
@Service
public class AccountServiceImpl implements AccountService {
       @Transactional
       public void test() {
       }
}
```

now at startup time, a new class is created, called proxy. This one is in charge of adding Transactional behavior as follows:
transactional-proxy-after-startup

Then when we invoke applicationContext.getBean(AccountService.class) we've got our proxy.

**Example 4** - Component with @Transactional :
```
@Component
public class AccountService {
     @Transactional
     public void create() {
     }
}
```

By default, if your bean does not implement an interface, Spring uses technical inheritance: at startup time, a new class is created. It inherits from your bean class and adds behavior in the child methods.

proxy-transaction-inheritance

### DIFF
--- a/platform/event/src/main/java/org/motechproject/event/listener/proxy/EventAnnotationBeanPostProcessor.java
+++ b/platform/event/src/main/java/org/motechproject/event/listener/proxy/EventAnnotationBeanPostProcessor.java
@@ -68,40 +68,38 @@ public class EventAnnotationBeanPostProcessor implements DestructionAwareBeanPos
         if (bean == null) {
             return;
         }
-        ReflectionUtils.doWithMethods(bean.getClass(), new ReflectionUtils.MethodCallback() {
 
-            @Override
-            public void doWith(Method method) throws IllegalAccessException {
-                Method methodOfOriginalClassIfProxied = ReflectionUtils.findMethod(AopUtils.getTargetClass(bean), method.getName(), method.getParameterTypes());
+        //Get declared methods from class without superclass methods.
+        for (Method method : bean.getClass().getDeclaredMethods()) {
+            Method methodOfOriginalClassIfProxied = ReflectionUtils.findMethod(AopUtils.getTargetClass(bean), method.getName(), method.getParameterTypes());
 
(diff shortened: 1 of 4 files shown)


kmadej:
I will add tests in another PR

PR #438, opened 2016-03-17
MOTECH-2306 Removed unnecessary data from instances requests

sebbrudzinski:
The endpoints for retrieving a number of instances for
grid display would return a huge number of data that is
unnecessary in grid display. New "basic" classes have been
implemented, that contain only required data and are now
used to pass data to view for instances requests.


platform/mds/mds-web/src/main/java/org/motechproject/mds/web/domain/EntityRecord.java
```diff
@@ -41,8 +29,14 @@ public void setEntitySchemaId(Long entitySchemaId) {
         this.entitySchemaId = entitySchemaId;
     }
 
-    public void setFields(List<FieldRecord> fields) {
-        this.fields = fields;
+    @JsonProperty("fields")
+    public List<FieldRecord> getFieldRecords() {
```
pgesek: why not override the getFields method?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -536,23 +539,33 @@ private void updateFields(Object instance, List<FieldRecord> fieldRecords, Motec
...
+                                                    MotechDataService service) {
+        return instanceToRecord(instance, entityDto, fields, service, true);
+    }
+
     private EntityRecord instanceToRecord(Object instance, EntityDto entityDto, List<FieldDto> fields,
                                           MotechDataService service) {
+        return (EntityRecord) instanceToRecord(instance, entityDto, fields, service, false);
+    }
+
+    private BasicEntityRecord instanceToRecord(Object instance, EntityDto entityDto, List<FieldDto> fields,
```
pgesek: Maybe instance passing a boolean, pass a class<? extends BasicEntityRecord), so that you don't have to keep casting the results

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/util/Constants.java
```diff
@@ -0,0 +1,10 @@
+package org.motechproject.mds.web.util;
+
+public final class Constants {
```
pgesek: 1) javadoc
2) do we need these constants? don't we already have these defined in mds?

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
sebbrudzinski: Pawel, can you elaborate on this point? How is passing the class reference supposed to help me here?

pgesek: make this a generic <T> method that returns T and takes a Class<T extends BasicRecord> parameter, that's what I meant. That way it will know what T is from the class parameter. Many methods in libraries do that

PR #443, opened 2016-03-18
MOTECH-2292: Fixed [MDS]  New Lookup field causes null pointer exception

pkornowski:
When we are creating lookup field, at the first lookupDTO's field takes a null name, then we are choosing the field name from the list, so that name will be overwritten. Null name gives exception, because of it is used in map.put(key, value) method, like key. If map is persistent then key can't be null. That's why I implemented the condition which catches null name at the beginning and replaces it with chosen name.


platform/mds/mds/src/main/java/org/motechproject/mds/domain/Entity.java
```diff
@@ -641,6 +641,9 @@ public void updateIndexes(List<LookupDto> indexes) {
                 }
                 String lookupFieldName = LookupName.buildLookupFieldName(field.getName(), lookupField.getRelatedName());
                 lookupFieldsOrder.add(lookupFieldName);
+                if(lookupField.getName() != field.getName()) {
```
pgesek: when does this happen?

pgesek: Moreover, you should not use != and == for comparing strings, see http://stackoverflow.com/questions/767372/java-string-equals-versus

pkornowski:
@pgesek  That 'if' condition is unnecessary, I'm sorry about that. 

I tested this with "Related Field" and it working.

platform/mds/mds/src/main/java/org/motechproject/mds/domain/Entity.java
```diff
@@ -641,6 +641,7 @@ public void updateIndexes(List<LookupDto> indexes) {
                 }
                 String lookupFieldName = LookupName.buildLookupFieldName(field.getName(), lookupField.getRelatedName());
                 lookupFieldsOrder.add(lookupFieldName);
+                lookupField.setName(lookupFieldName);
```
pgesek: Is there any test that we can extend to catch this?

platform/mds/mds/src/main/java/org/motechproject/mds/dto/DraftData.java
```diff
@@ -21,6 +21,7 @@
     public static final String FIELD = "field";
 
     public static final String ADD_NEW_INDEX = "$addNewIndex";
+    public static final String ADD_NEW_FIELD = "indexes.0.$addField";
```
kmadej: do not add constant with hardcoded path for adding new field to the first lookup from the list. I see that you using this only in test, just use this as a raw string in test.

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
 
         entityService.saveDraftEntityChanges(entityDto.getId(), draftData);
+
         entityService.commitChanges(entityDto.getId());
 
         assertTrue(containsLookup("Lookup 1", entityDto.getId()));
     }
 
     @Test
+    public void shouldAddNewFieldForLookupAndSaveEntity() throws IOException {
```
kmadej: please remove any redundant line from this test

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        values.put("path", DraftData.ADD_NEW_FIELD);
+        values.put("advanced", true);
+
+        DraftData draftData = new DraftData();
+        draftData.setEdit(true);
+        draftData.setValues(values);
+        entityDto = entityService.createEntity(entityDto);
+
+        FieldDto testField = FieldTestHelper.fieldDto("firstField", Boolean.class);
+        entityService.addFields(entityDto, asList(testField));
```
kmadej: do not use asList with only one field - use singletonList

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        List<LookupFieldDto> lookupFieldDtos = lookupFieldsFromNames("firstField");
+        LookupDto lookup = new LookupDto("lookup", false, false, lookupFieldDtos, true);
+        entityService.addLookups(entityDto.getId(), asList(lookup));
+
+        entityService.saveDraftEntityChanges(entityDto.getId(),
+                DraftBuilder.forNewField("disp", "f1name", Long.class.getName()));
+        entityService.getFields(entityDto.getId());
```
kmadej: i think this line is redundant

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        FieldDto field = selectFirst(entityService.getFields(entityDto.getId()),
+                having(on(FieldDto.class).getBasic().getName(), equalTo("f1name")));
+
+        values.put("value", asList(field.getId()));
+        entityService.saveDraftEntityChanges(entityDto.getId(), draftData);
```
kmadej: why you use saveDraftEntityChanges twice in this test ?

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        entityService.commitChanges(entityDto.getId());
+
+        assertEquals(entityService.getLookupByName(entityDto.getId(),"lookup").getLookupField("f1name").getName(), "f1name");
```
kmadej: expected value should be on the left side

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -209,12 +209,48 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        entityDto = entityService.createEntity(entityDto);
+
+        FieldDto testField = FieldTestHelper.fieldDto("firstField", Boolean.class);
+        entityService.addFields(entityDto, asList(testField));
+        List<LookupFieldDto> lookupFieldDtos = lookupFieldsFromNames("firstField");
+        LookupDto lookup = new LookupDto("lookup", false, false, lookupFieldDtos, true);
+        entityService.addLookups(entityDto.getId(), asList(lookup));
+
+        entityService.saveDraftEntityChanges(entityDto.getId(),
+                DraftBuilder.forNewField("disp", "f1name", Long.class.getName()));
```
kmadej: btw maybe change f1name to for example fieldName

pkornowski:
@kmadej There is saveDraftEntityChanges used twice, because field which we want to add into lookup need to be in draftData before editing Lookup. First saveDraftEntityChanges create field in draftData which we are using after.

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,36 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        DraftData draftData = new DraftData();
+        draftData.setEdit(true);
+        draftData.setValues(values);
+        entityDto = entityService.createEntity(entityDto);
+
+        FieldDto testField = FieldTestHelper.fieldDto("firstField", Boolean.class);
+        entityService.addFields(entityDto, Collections.singletonList(testField));
+        List<LookupFieldDto> lookupFieldDtos = lookupFieldsFromNames("firstField");
+        LookupDto lookup = new LookupDto("lookup", false, false, lookupFieldDtos, true);
+        entityService.addLookups(entityDto.getId(), Collections.singletonList(lookup));
```
kmadej: you should not add to entity, work on draft, then commitChanges

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,36 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        entityService.saveDraftEntityChanges(entityDto.getId(), DraftBuilder.forNewField("disp", "f1name", Long.class.getName()));
+        entityService.getFields(entityDto.getId());
```
kmadej: redundand

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,36 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        values.put("path", "indexes.0.$addField");
+        values.put("advanced", true);
+        DraftData draftData = new DraftData();
+        draftData.setEdit(true);
+        draftData.setValues(values);
+        entityDto = entityService.createEntity(entityDto);
+
+        FieldDto testField = FieldTestHelper.fieldDto("firstField", Boolean.class);
+        entityService.addFields(entityDto, Collections.singletonList(testField));
+        List<LookupFieldDto> lookupFieldDtos = lookupFieldsFromNames("firstField");
```
kmadej: why adding here lookupField ? Just testing adding this without any field in lookup

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,36 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        EntityDto entityDto = new EntityDto();
+        entityDto.setName("myEntity");
+
+        Map<String, Object> values = new HashMap<>();
+        values.put("path", "indexes.0.$addField");
+        values.put("advanced", true);
+        DraftData draftData = new DraftData();
+        draftData.setEdit(true);
+        draftData.setValues(values);
+        entityDto = entityService.createEntity(entityDto);
```
kmadej: please group lines in this test, it's messy. This line with creating new entity should be above creating drafdata

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,32 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+    public void shouldAddNewFieldForLookupAndSaveEntity() throws IOException {
+        EntityDto entityDto = new EntityDto();
+        entityDto.setName("myEntity");
+        entityDto = entityService.createEntity(entityDto);
+
+        EntityDraft entityDraft = entityService.getEntityDraft(entityDto.getId());
+        LookupDto lookup = new LookupDto("lookup", false, false, null, true);
+        entityService.addLookups(entityDraft.getId(), Collections.singletonList(lookup));
+        entityService.saveDraftEntityChanges(entityDto.getId(), DraftBuilder.forNewField("disp", "testFieldName", Long.class.getName()));
+        FieldDto field = selectFirst(entityService.getFields(entityDto.getId()), having(on(FieldDto.class).getBasic().getName(), equalTo("testFieldName")));
```
kmadej: group this once again, after creating an entity, create field and then start with lookup

platform/mds/mds/src/test/java/org/motechproject/mds/it/service/EntityServiceContextIT.java
```diff
@@ -215,6 +215,32 @@ public void shouldSaveEntityWithGivenLookups() throws IOException {
...
+        values.put("advanced", true);
+        values.put("value", Collections.singletonList(field.getId()));
+        DraftData draftData = new DraftData();
+        draftData.setEdit(true);
+        draftData.setValues(values);
+
+        entityService.saveDraftEntityChanges(entityDto.getId(), draftData);
+        entityService.commitChanges(entityDto.getId());
+
+        assertEquals("testFieldName", entityService.getLookupByName(entityDto.getId(),"lookup").getLookupField("testFieldName").getName());
```
kmadej: this assert is wrong because when this field doesn't exist it will be NPE here, please write assert in different way

kmadej:
looks good to me, please cherry-pick to 0.28.X branch

PR #445, opened 2016-03-21
MOTECH-2306 Removed unnecessary data from instances requests 

sebbrudzinski:
The endpoints for retrieving a number of instances for
grid display would return a huge number of data that is
unnecessary in grid display. New "basic" classes have been
implemented, that contain only required data and are now
used to pass data to view for instances requests.


sebbrudzinski:
test this please

sebbrudzinski:
test this please

PR #446, opened 2016-03-21
MOTECH-2316: Eliminated reliance on provider ID in tasks

pgesek:
Since the ID could change between instances, the current state of affairs would prevent importing/exporting (unless you are lucky) and potentially cause issues on restarts. I have eliminated this dependency by making tasks depend on provider names (now validated).


pgesek:
I will look into compatiblity with older task versions

PR #448, opened 2016-03-22
MOTECH-2054 Replace LoadOnDemand.js

wstrzelczyk:


sebbrudzinski:
Will resubmit when it's ready for review

PR #450, opened 2016-03-22
MOTECH-2319: Unschedules task retry when task is disabled

kmadej:


modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskTriggerHandlerTest.java
```diff
@@ -1611,7 +1610,7 @@ public void shouldUnScheduleTaskRetriesWhenTaskDeleted() throws Exception {
     }
 
     @Test
-    public void shouldNotHandleTaskRetryWhenTaskDisabled() throws Exception {
+    public void shouldUnScheduleTaskRetriesWhenTaskDisabled() throws Exception {
```
sebbrudzinski: Unschedule

PR #452, opened 2016-03-23
MOTECH-2317: Retry count for task failure has higher priority than possible errors in settings

kmadej:
Now retry count for task failure has higher priority than possible errors
number in settings.

I added an unit test and remove one redundant line from TaskTriggerHandlerTest.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskTriggerHandler.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskTriggerHandler.java
@@ -240,7 +240,8 @@ public class TaskTriggerHandler implements TriggerHandler {
         int failureNumber = task.getFailuresInRow();
         int possibleErrorsNumber = getPossibleErrorsNumber();
 
-        if (failureNumber >= possibleErrorsNumber) {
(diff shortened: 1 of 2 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskTriggerHandler.java
```diff
@@ -240,7 +240,8 @@ private void handleError(Map<String, Object> params, Task task, TaskHandlerExcep
         int failureNumber = task.getFailuresInRow();
         int possibleErrorsNumber = getPossibleErrorsNumber();
 
-        if (failureNumber >= possibleErrorsNumber) {
+        // Retry count for task failure has higher priority than possible errors number in settings
+        if (!task.retryTaskOnFailure() && failureNumber >= possibleErrorsNumber) {
```
pgesek: hm, so these tasks will never disable? wondering if that's what we want

pgesek:
I gave this some thought and I believe it makes most sense to also control this setting on a per task level. Moreover, the retry and retry interval settings should also have their global setting counterparts - indicated in the task form UI. @kmadej can you bring this up to the dev list? I am closing this PR for now

PR #453, opened 2016-03-23
Added missing modules to repository install

sebbrudzinski:
Four missing modules have been added to the dropdown list for install
from repository. Moreover, the install button will no longer work if
no module or file was selected.


sebbrudzinski:
test this please

PR #454, opened 2016-03-23
[0.27.X] MOTECH-2316: Switching provider IDs used in tasks to provider names

pgesek:
Using IDs was a wrong design decision here, they cause a plethora of issues. Compatibility for old tasks is retained thanks to the introduction of a task migrator mechanism. Moreover, a fix for tasks event listeners not being unregistered was also introduced.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/ProviderNotFoundException.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/ProviderNotFoundException.java
@@ -5,7 +5,9 @@ package org.motechproject.tasks.ex;
  */
 public class ProviderNotFoundException extends IllegalArgumentException {
 
-    public ProviderNotFoundException(long providerId) {
-        super("Unable to migrate task. Data provider with id " + providerId + " not found");
+    private static final long serialVersionUID = -5419738008480651643L;
+
+    public ProviderNotFoundException(String taskName, long providerId) {
+        super("Unable to migrate task: " + taskName + ". Data provider with id " + providerId + " not found");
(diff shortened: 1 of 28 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/ProviderNotFoundException.java
```diff
@@ -0,0 +1,11 @@
+package org.motechproject.tasks.ex;
+
+/**
+ * Signals that provider with a given id was not found.
+ */
+public class ProviderNotFoundException extends IllegalArgumentException {
+
+    public ProviderNotFoundException(long providerId) {
+        super("Unable to migrate task. Data provider with id " + providerId + " not found");
```
sebbrudzinski: Including the task name for which the migration has failed would be helpful as well.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/ex/ProviderNotFoundException.java
```diff
@@ -0,0 +1,11 @@
+package org.motechproject.tasks.ex;
+
+/**
+ * Signals that provider with a given id was not found.
+ */
+public class ProviderNotFoundException extends IllegalArgumentException {
+
```
sebbrudzinski: serialVersionUID

pgesek: I doubt we will be serializing this, but I guess I can add it for the sake of removing an IDE warning

PR #455, opened 2016-03-23
MOTECH-2265: period modal directive should work after uglification

1stmateusz:


kmadej:
@1stmateusz please write in the topic of the PR what you did. After your changes period modal directive should work, so the topic is confusing

nickdotreid:
@1stmateusz 

Since we are refactoring a little, can we change this implementation to use the BootstrapDialog modal, rather than just Bootstrap's plain modal implementation? Ideally the entire codebase should be moving in this direction....

This would require removing everything but the #period-slider element from [the period-modal template](https://github.com/1stmateusz/motech/blob/motech-2265/platform/server-bundle/src/main/resources/webapp/partials/period-modal.html) which should be named something more semantic

also, @pgesek @llavoie please tell me if I'm overstepping the scope of this ticket...

pgesek:
@nickdotreid from my understanding this ticket is about making the period widget work after uglification - on the current master it only works before uglification. The only reason I brought this ticket back from Post-1.0 is because the same issue affects the new file-input widget and I want to get that whole thing closed. We can do refactoring in a separate PR/ticket after the uglification thing gets sorted.

nickdotreid:
Fair enough  ignore my comment

1stmateusz:
It appears that Angular applications are not min-safe in each case. The code written in period modal directive was not min-safe at all. The problem occur when uglify plugin rename variables like "$http" into "a" for example. It is required to use array annotation to prevent renaming of these variables. Instead of using `directives.directive('periodModal', function($http, $templateCache, $compile)` we need to insert strings containing names `widgetModule.directive('periodModal', ['$compile', '$http', '$templateCache',function($http, $templateCache, $compile)`.
Moreover writing `directives.directive('periodModalMds', function($http, $templateCache, $compile) {
return function(scope, element, attrs) {` (placing "return function()" at the start of the diretive) also appears to be not min-safe. Functions inside return statement should be assigned to some name, for example `compile: function(){`. 
This information was found in [min-safe](http://www.sitepoint.com/5-minutes-to-min-safe-angular-code-with-grunt/)

kmadej:
@1stmateusz please remember cherr-pick to 0.28.X

PR #456, opened 2016-03-24
[0.28.X] MOTECH-2316: Switching provider IDs used in tasks to provider names

pgesek:


sebbrudzinski:
@pgesek Shouldn't MOTECH Commons provider be renamed here as well?

sebbrudzinski:
The provider has been renamed, therefore I've merged the pull request now.

PR #457, opened 2016-03-24
MOTECH-2208: Only one modal on a screen at a time

akarczewski:


akarczewski:
related with motech/modules#221

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1566,34 +1566,51 @@
...
-                     unblockUI();
+                     ModalService.unblockUI();
                  }, function (response) {
-                     unblockUI();
-                     handleResponse('task.header.error', 'task.history.deleteError', response);
+                     ModalService.unblockUI();
+                     ModalService.handleResponse('task.header.error', 'task.history.deleteError', response);
                  });
             });
+//            ModalService.confirm2 ({
```
sebbrudzinski: Remove commented out code.

platform/server/src/main/webapp/static/js/bootstrap-page.js
```diff
@@ -234,5 +234,22 @@ function retrieveStatus() {
...
 function submitForm() {
-    blockUI();
+    var dialog = new BootstrapDialog({
+        message: function(dialogRef){
+            var $message = $(
+                '<div class="splash-logo"><img src="./../../static/common/img/motech-logo.gif" alt="motech-logo"></div>' +
+                '<div class="clearfix"></div>' +
+                '<div class="splash-loader"><img src="./../../static/common/img/loadingbar.gif" alt="Loading..."></div>' +
+                '<div class="clearfix"></div>' + '<br>');
+                return $message;
```
nickdotreid: Lets get this HTML markup into a an actual HTML file somewhere, and out of the javascript

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -7,4 +7,269 @@
...
+                        if (typeof dialog.getData('callback') === 'function' && dialog.getData('callback').call(this, true) === false) {
+                            return false;
+                        }
+                    }
+                }]
+            });
+            this.open(dialog);
+        };
+
+        this.motechAlert = function (msg, title, params, callback) {
```
nickdotreid: A better function name would be alert

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -7,4 +7,269 @@
...
+            this.alert({
+                title: jQuery.i18n.prop(title),
+                message: jQuery.i18n.prop.apply(null, [msg].concat(params)),
+                callback: callback
+            });
+        };
+
+        this.motechConfirm = function (msg, title, callback) {
```
nickdotreid: A better function name would be confirm

nickdotreid: A better function name would be confirm

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -7,4 +7,269 @@
...
+            this.confirm({
+                title: jQuery.i18n.prop(title),
+                message: jQuery.i18n.prop(msg),
+                callback: callback
+            });
+        };
+
+        this.motechAlertStackTrace = function (msg, title, response, callback) {
+            if( title === null || title === '') {
```
nickdotreid: Does it make sense to set default titles on other Alerts and Confirms?

Also, this string needs to be translated

nickdotreid:
@akarczewski -- great start...

Can we refactor the ModalService into a couple smaller Services, I think this would clean up the OO design and everything easier to read... I'd suggest:

- BootstrapDialog
Which would wrap the BootstrapDialog service with set defaults and deal with window overlapping...

- LoadingModal
Which has methods 'open', 'close', and 'isOpen' and would take the place of 'blockUI' and 'unblockUI' -- you could also call this service 'BlockUI' but the blockUI is typically a representation we use when MOTECH is loading something and eventually it should show state of what it's loading... so if we are going to make big changes, might as well get the verbs correct... 

- Modal
This is your ModalService, and it would call both BootstrapDialog and LoadingModal. This would be the service that other controllers would implement, not BootstrapDialog, so we can call an alert with 'alert' rather than 'motechAlert' (same with confirm).

Having this class inherit LoadingModal means we can also check if LoadingModal.isOpen() and LoadingModal.close() it if needed when we fire an alert or confirm, which means we can remove functions like alertHandler that have unclear functionality of unblocking the UI. Honestly, I'm not 100% sure I like Modal.alert() closing a LoadingModal... I feel like that is something the for the controller to do... but I'm not sold on this yet .... 

Also, functions like alertHandlerWithCallback shouldn't exist... alertHandler should have an optional argument thats a callback.... and ideally you would use $q and have alert and confirm return a deferred promise, so the end code would read:
`Modal.alert('whatever').then(function); Modal.confirm('cheese with that').then(function).catch(otherFunction);`

Sidenote:
You might have also noticed I didn't call any of these services a 'Service' -- because generally a service connotes that it relates to a data source, and these actually control the UI -- even though technically we need the singleton aspect of the service... Angular libraries like Ionic do this, and the code feels slick to write, so I don't feel bad copying them...

platform/server-bundle/src/main/resources/webapp/js/services.js
akarczewski: The default title is "Success" on other Alerts and Confirms. We can always set other when needed.

nickdotreid: @akarczewski This might be a bad idea (so feel free to push back), but would it make sense to build the default titles into the default calls? (meaning this.alert or this.confirm)

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,277 @@
     'use strict';
 
     var uiServices = angular.module('uiServices', ['ngResource']);
-
     uiServices.factory('Menu', function($resource) {
         return $resource('module/menu');
     });
 
+    uiServices.service('BootstrapDialogManager', function () {
+        var modals2 = [];
```
nickdotreid: Can we come up with a more descriptive variable name then 'modals2'?

Suggestions might include:
'modals'
'modalsList'
'modalWindows'
'modalWindowStack'

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,277 @@
...
+            });
+        };
+
+        this.motechAlertStackTrace = function (msg, title, response, callback) {
+            if( title === null || title === '') {
+                title = 'Alert';
+            }
+            this.alert({
+                title: title,
+                message: jQuery.i18n.prop(msg).bold() + ": \n" + response,
```
nickdotreid: Let's not put formatting into the code directly. 

Can we use $compile here and render a template? This would allow us to use a directive (or something) to format the stack trace (in the future eventually) without touching the javascript...

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,277 @@
...
+            this.alert({
+                title: jQuery.i18n.prop(title),
+                message: jQuery.i18n.prop.apply(null, [msg].concat(params)),
+                callback: callback
+            });
+        };
+
+        this.motechAlertStackTrace = function (msg, title, response, callback) {
+            if( title === null || title === '') {
+                title = 'Alert';
```
nickdotreid: Use a translatable string

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,277 @@
...
+        };
+
+        this.isOpen = function () {
+            return open;
+        };
+    });
+
+    uiServices.service('Modal', function (LoadingModal, BootstrapDialogManager) {
+
+        this.alert = function () {
```
nickdotreid: I got really confused that this function didn't take any arguments, and then sets them further down in the code...could we move the defaultOptions out of the function to expose what you are doing with the arguments array more immediately?

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,277 @@
...
+            }
+
+            dialog = new BootstrapDialog({
+                type: options.type,
+                title: options.title,
+                message: options.message,
+                closable: options.closable,
+                draggable: options.draggable,
+                data: {
+                    callback: options.callback
```
nickdotreid: I don't think the callback function belongs here

nickdotreid:
@akarczewski 

**LoadingModal**
I completely take back what I said in allowing the Modal service to use LoadingModal. These services should be completely separate and not have knowledge of each other...

I think its more clear in a controller when a _LoadingModal_ is _.open()_ed and then _.close()_ed instead of _Modal.openLoadingModal()_ then _Modal.alertHandle()_ (which closes the LoadingModal) -- this is more clear, even though this does add additional lines to the controller (one to close the loading modal, and another to show the alert). I also like _LoadingModal.open()_ more than _Modal.openLoadingModal()_ and feel that it's good form for a controller to state that it uses both a Modal and a LoadingModal

**Modal Service**
Your implementation of Modal.motechAlert is what I expected Modal.alert to do... and I think ideally we should be exposing more readable function names to controller code...

Can we change this.alert to a function within the scope of the service called _alert_ or _makeAlert_ because this is a private function that should not be callable outside the service.

I also think it would make sense to make the service Modal a factory, and call it ModalFactory, where each function passes back a BootstrapDialog instance, which can then be closed and manipulated in the controller

I don't like the *handler functions in this service, I feel that creating a lambda is something that can be done more clearly in the controller directly (or at least in the instances I've noticed in your code)

**BootstrapDialogManager**
I like your BootstrapModalManager service. I think it is clean. I don't love that we need to include the BootstrapDialogManger into both Modal and LoadingModal services. I'd rather use event listeners to listen for bootstrapModal events (which I think BootstrapDialog publishes) or completely abstract BootstrapDialog, such that Modal and LoadingModal don't need to worry about reaching out to BootstrapDialogManager (or even know it exists).

I'm ok with not fully abstracting BootstrapDialog... but can we wrap it as an Angular constant, so that AngularJS will break at runtime if BootstrapDialog isn't loaded (instead of when we call a Modal) and that we can tell which functions use BootstrapDialog

wstrzelczyk:
The new changes have been implemented in the new pull request:
https://github.com/motech/motech/pull/490

sebbrudzinski:
@nickdotreid @wstrzelczyk 
I'm closing this pull request. Please review/comment in the new pull request, linked by Waldek.

PR #459, opened 2016-03-25
MOTECH-2298: Changed the way BaseListeners retrieve application context [0.28.X]

ngraczewski:
MDS BaseListeners will now try to retrieve entities bundle application context from the bundle context before waiting for the notification.


ngraczewski:
Branches mismatch

PR #472, opened 2016-03-30
MOTECH-2124 Modifed Authentication Handler for JSON Login

nickdotreid:
Added logic to authentication handlers that catches XMLHttpRequests, and
returns text, instead of redirecting. This was created to support a
seperated UI that would communicate with MOTECH via JSON.


platform/web-security/src/main/java/org/motechproject/security/authentication/MotechLoginErrorJSONHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.security.authentication.LockedException;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.authentication.ExceptionMappingAuthenticationFailureHandler;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class MotechLoginErrorJSONHandler extends ExceptionMappingAuthenticationFailureHandler {
```
pgesek: Is it just me, or does it return plaintext, not json?

platform/web-security/src/test/java/org/motechproject/security/authentication/MotechLoginSuccessHandlerTest.java
```diff
@@ -84,4 +86,22 @@ public void shouldResetFailureLoginCounter() throws ServletException, IOExceptio
...
+        MotechUser user = new MotechUser();
+        user.setUserName("testUser");
+        user.setFailureLoginCounter(3);
+
+        when(authentication.getName()).thenReturn("testUser");
+        when(motechUsersDao.findByUserName("testUser")).thenReturn(user);
+
+        motechLoginSuccessHandler.onAuthenticationSuccess(mockRequest, mockResponse, authentication);
+
+        assertEquals("SUCCESS", mockResponse.getContentAsString());
```
pgesek: doesn't look like json

nickdotreid:
@pgesek right now its not returning actual JSON, there are plenty of other APIs in MOTECH that do this, which I believe is bad...

I'll use gson to build a JSON Object, but I'm curious if it makes sense to build a reusable class for passing back messsages (and if we did do that, where should that class live?)

I'll change that and get it submitted...

pgesek:
Don't think we do it in that many spots. I believe we have a status endpoint somewhere that does this, but I guess that one is more about the http code anyway (we should switch it to json for consistency). Perhaps also some UI endpoints for the dashboard that return the stuff like server, guess those should get switched to json for consistency as well.  

Not entirely sure what you mean with the common class, what would this class do? Do you mean a json object that just has one string value?

nickdotreid:
Right now I'm imagining in most cases we would return a JSON object that
would look like this:
*{*
*    "message": "Foo Bar"*
*}*

I know that in most cases the @ResponseBody in SpringMVC is supposed to
deal with a format objects that are passed back... so having a simple class
that gets instantiated with a String argument called "message" would help
build some consistency going forward.

I've started making a list of API endpoints exposed in MOTECH, with the
hope that after making the list we can have a conversation about making
everything more consistent...
https://docs.google.com/spreadsheets/d/12p3I_EEvUiy-SeaD1HmgbWNRuNWWR0g9e7Qk08mmYzI/edit?usp=sharing

-- nick --

pgesek:
I left some comments on that doc. The perfect place for such a class would be the commons module, although as I mentioned in a comment on the doc, it is worth considering grouping some of the requests together.

nickdotreid:
@pgesek this now returns solid JSON....

I was thinking of adding a method to MotechJsonMessage which would look like:
writeJson(HttpServletResponse response)

This would take a response, and write the Json response and adjust headers to Content-Type: application/json // but I felt this was an over reach.... thoughts?

pgesek:
@nickdotreid I think it's fine as it is. In most cases we should just Spring handle jsonification though

PR #473, opened 2016-03-31
MOTECH-2292: Fixed problem with double click and adding more lookups field than available

pkornowski:
Corrections after test:

- When we click on "New Lookup Field" button, then this button will be blocked until the field will be added;

- When we click on "New Lookup" button, then this button will be blocked until the lookup will be added;

- When there is no available lookup field, then button New Lookup Field  will be blocked;


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -877,6 +877,7 @@
             $scope.advancedSettings = Entities.getAdvanced({id: $scope.selectedEntity.id},
                 function () {
                     $scope.blockLookups = false;
+                    $scope.notDisabled = true;
```
kmadej: 1. this is not the best name, in boolean fields you shouldn't use 'not', I need few seconds to understand what true means for this field
2. apart from 1., the name is not adequate to purpose of this field.

plus i think there should be two boolean fields for add new lookup button and for add new lookup field button. Because now when for example 'new lookup field' button is disabled, the button 'new lookup' is also disabled.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2001,24 +2002,28 @@
         * Adds a new index and sets it as the active one
         */
         $scope.addNewIndex = function () {
-            var newLookup = {
-                lookupName: $scope.getLookupNameForNewLookup(),
-                singleObjectReturn: true,
-                indexRequired: true,
-                lookupFields: []
-            };
+            if($scope.notDisabled === true) {
```
kmadej: 1. if ($scope.notDisabled)
2. why you need this condition ?  is there a possibilty that it will be false here ?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2133,23 +2138,28 @@
         /**
         * Adds new lookup field to the currently active index
         */
+
```
kmadej: remove this line

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2133,23 +2138,28 @@
...
         $scope.addLookupField = function () {
-            var value = $scope.availableFields[0] && $scope.availableFields[0].id;
+            if($scope.notDisabled === true) {
```
kmadej: the same as above

platform/mds/mds-web/src/main/resources/webapp/partials/modals/partials/indexesLookups.html
```diff
@@ -10,7 +10,7 @@
                 </div>
             </div>
             <div class="form-group">
-                <button ng-click="addNewIndex()" class="btn btn-success" ng-disabled="blockLookups" type="button">
+                <button ng-click="addNewIndex()" class="btn btn-success" ng-disabled="blockLookups || notDisabled === false" type="button">
```
kmadej: instead of condition notDisabled === false you should use !notDisabled

platform/mds/mds-web/src/main/resources/webapp/partials/modals/partials/indexesLookups.html
```diff
@@ -132,7 +132,7 @@
 
             <div class="form-group">
                 <div class="margin-before2">
-                    <button type="button" class="btn btn-success" ng-click="addLookupField()" ng-show="!lookup.readOnly && !lookup.referenced" ng-disabled="lookup.referenced">
+                    <button type="button" class="btn btn-success" ng-click="addLookupField()" ng-show="!lookup.readOnly && !lookup.referenced" ng-disabled="lookup.referenced || notDisabled === false || availableFields.length === 0">
```
kmadej: the same as above

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2018,7 +2021,9 @@
             }, function () {
                 $scope.advancedSettings.indexes.push(newLookup);
                 $scope.setActiveIndex($scope.advancedSettings.indexes.length-1);
+                $scope.isNewLookupButtonDisabled = false;
```
pgesek: We should also reenable the buttons in case of a server-side error. The draft function currently doesn't support this, but we should extend it with an errorCallback parameter. It should be executed at the end of the errorHandler which the draft() function is currently using.

pkornowski:
test this please

kmadej:
Add new parameter errorCallback to the draft function, this callback should be executed at the end of errorHandler in draft function. Then invoke this updated draft function with your own errorCallback, which should enable this buttons

kmadej:
@pkornowski please cherry-pick this to 0.28.X

PR #476, opened 2016-03-31
MOTECH-2322: Fixed: Instance restored from trash remains in trash

pkornowski:
If instance will be restored, then object in trash will be removed thanks for moveFromTrash function. I added integration test for reverting from trash too.


platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java
```diff
@@ -317,8 +317,11 @@ public T revertFromTrash(Long trashId) {
             T newInstance = getClassType().newInstance();
 
             copyValuesFromRecord(newInstance, trashRecord);
+            newInstance = repository.create(newInstance);
```
kmadej: use create from motechdataservice, after your change we lost sending create event

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/domain/RevertFromTrash.java
```diff
@@ -0,0 +1,32 @@
+package org.motechproject.mds.test.domain;
+
+import org.motechproject.mds.annotations.Entity;
+import org.motechproject.mds.annotations.Field;
+
+@Entity(recordHistory = true)
+public class RevertFromTrash {
```
kmadej: change the name of test class. This class should be in package domain.revertFromTrash, however the name should be normal, look for the entity names in other tests

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/service/TestRevertFromTrashService.java
```diff
@@ -0,0 +1,8 @@
+package org.motechproject.mds.test.service;
+
+import org.motechproject.mds.service.MotechDataService;
+import org.motechproject.mds.test.domain.RevertFromTrash;
+
+
+public interface TestRevertFromTrashService extends MotechDataService<RevertFromTrash> {
```
kmadej: this service also should be in the new package and add osgi reference for this service

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -1984,6 +1994,18 @@ public void shouldCreateHistoryForManyToMany() {
         assertEquals(jackDt2, safeGetProperty(secondVersion, "modificationDate"));
     }
 
+    @Test
+    public void shouldRevertFromTrash() {
+        RevertFromTrash testRevert = testRevertFromTrashService.create(new RevertFromTrash("revertTest"));
```
kmadej: this test should look differently, try to test the real use case. First of all after refactor trash and history to use instance lifecycle listeners, an instance should be moved to trash while deleting instance. So do not use method moveToTrash, just delete the instance.

you always use id of created instance, in this example there are only one instance in main entity and one in trash, so the id probably is the same and has value 1. I think this test should be a bit more complicated. Can you create two instances,  delete them, then revert one of them from trash and check at the end if there is one instance in main table and one in trash

platform/mds/mds-test-bundle/src/main/java/org/motechproject/mds/test/service/TestRevertFromTrashService.java
```diff
@@ -0,0 +1,8 @@
+package org.motechproject.mds.test.service;
+
+import org.motechproject.mds.service.MotechDataService;
+import org.motechproject.mds.test.domain.RevertFromTrash;
+
+
```
kmadej: remove redundant line

platform/mds/mds-test-bundle/pom.xml
```diff
@@ -81,7 +82,8 @@
                             org.motechproject.mds.test.service.lookupcomboboxrelation;version=${project.version},
                             org.motechproject.mds.test.service.editablelookups;version=${project.version},
                             org.motechproject.mds.test.service.mapdeserialisation;version=${project.version},
-                            org.motechproject.mds.test.service.optimisticlocking;version=${project.version}
+                            org.motechproject.mds.test.service.optimisticlocking;version=${project.version},
+                            org.motechproject.mds.test.service.restoreInstanceFromTrash;version=${project.version}
```
pgesek: why are the packages for service and domain inconsistent?

platform/mds/mds/src/main/java/org/motechproject/mds/service/DefaultMotechDataService.java
```diff
@@ -317,8 +317,11 @@ public T revertFromTrash(Long trashId) {
             T newInstance = getClassType().newInstance();
 
             copyValuesFromRecord(newInstance, trashRecord);
+            newInstance = create(newInstance);
 
-            return create(newInstance);
+            trashService.moveFromTrash(newInstance, trashRecord, recordHistory);
```
pgesek: I've taken a look at the implementation of this method and it makes no sense. Can you make the following changes:

* remove the newInstance and recordHistory params, since they are unused
* rename the method to removeFromTrash

PR #479, opened 2016-04-01
MOTECH-2306: Optimized a few methods in the InstanceServiceImpl class

ngraczewski:
The fetch time of 100 instances of the Participant went down from ~9 sec. to ~0.2 sec.


ngraczewski:
test this please

PR #480, opened 2016-04-01
Update Spring version to 3.2

pgesek:


platform/osgi-platform/src/main/resources/META-INF/services/javax.validation.spi.ValidationProvider
```diff
@@ -0,0 +1 @@
+org.apache.bval.jsr303.ApacheValidationProvider
```
sebbrudzinski: Since this is now automatically imported from every bundle, can we drop this file from MDS or should we keep it there for whatever reason?

sebbrudzinski:
@pgesek Looks good to me in general (see just one comment), but we may need to hold this off until changes for modules are ready to merge / conflicts are resolved to ensure build stability

pgesek:
We can leave the MDS imports explicit for the sake of them serving as an example

PR #486, opened 2016-04-07
MOTECH-2338: Timer class from metrics module moved to commons libraries

weronika-ciecierska:


motech-gerrit:

motech-gerrit:

motech-gerrit:

sebbrudzinski:
1. Remove unrealted commit from this branch (MOTECH-2232)
2. Change pull request title to something more descriptive

sebbrudzinski:
One more thing that we may want to do, is dropping the "TimerService" interface and "TimerServiceImpl", since all that they do is returning an instance of "Timer". Let's just expose the Timer class and drop unnecessary boilerplate, which is this service interfce and service implementation. This also means that we do not need to keep the blueprint file with service definition.

sebbrudzinski:
Changes will be address in a new pull request due to a mess with unrelated commits

PR #489, opened 2016-04-08
Updated the module list in admin

pgesek:
* motech-metrics -> metrics
* added ihe-interop

### DIFF
--- a/modules/admin/src/main/resources/webapp/js/controllers.js
+++ b/modules/admin/src/main/resources/webapp/js/controllers.js
@@ -118,9 +118,10 @@
         $scope.modules[$scope.mavenStr('dhis2')] = 'DHIS2';
         $scope.modules[$scope.mavenStr('event-logging')] = 'Event Logging';
         $scope.modules[$scope.mavenStr('http-agent')] = 'Http Agent';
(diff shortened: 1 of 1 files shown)


modules/admin/src/main/resources/webapp/js/controllers.js
```diff
@@ -118,9 +118,10 @@
         $scope.modules[$scope.mavenStr('dhis2')] = 'DHIS2';
         $scope.modules[$scope.mavenStr('event-logging')] = 'Event Logging';
         $scope.modules[$scope.mavenStr('http-agent')] = 'Http Agent';
+        $scope.modules[$scope.mavenStr('ihe-interop')] = 'IHE Interop';
         $scope.modules[$scope.mavenStr('ivr')] = 'IVR';
         $scope.modules[$scope.mavenStr('message-campaign')] = 'Message Campaign';
-        $scope.modules[$scope.mavenStr('motech-metrics')] = 'Metrics';
+        $scope.modules[$scope.mavenStr('motech')] = 'Metrics';
```
sebbrudzinski: metrics

pgesek: too fast for me

PR #490, opened 2016-04-11
MOTECH-2208 Only one modal on a screen at a time

wstrzelczyk:
MOTECH-2208 Enhanced modal window services


sebbrudzinski:
Changes for modules repository in: https://github.com/motech/modules/pull/242

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -111,29 +111,29 @@
...
+                        type: 'type-danger',
                         title: $scope.msg('task.error.actionNotChangeTitle'),
                         message: $scope.util.createErrorMessage($scope, response, false)
                     });
                 });
         };
 
         $scope.deleteTask = function (item) {
-            BootstrapDialog.confirm({
+            ModalFactory.confirm({
```
nickdotreid: Why doesn't this line use motechConfirm?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4182,15 +4182,15 @@
...
             History.getHistory({
                 entityId: $scope.selectedEntity.id,
                 instanceId: instanceId
             }, function () {
-                unblockUI();
+                LoadingModal.close();
                 $scope.previousInstance = undefined;
                 $scope.instanceId = instanceId;
-            }, angularHandler('mds.error', 'mds.error.historyRetrievalError'));
+            }, ModalFactory.angularHandler('mds.error', 'mds.error.historyRetrievalError'));
```
nickdotreid: I still don't see the need for the angularHandler function... I think it might be more clear to handle the error, and then use the familiar ModalFactory.motechAlert() function

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,274 @@
...
+
+    uiServices.service('LoadingModal', function () {
+        var dialog, open = false;
+
+        this.open = function () {
+            if (!open) {
+                dialog = new BootstrapDialog({
+                    message: function(dialogRef) {
+                        var $message = $('<div></div>'),
+                        pageToLoad = dialog.getData('pageToLoad');
```
nickdotreid: This seems like a really long winded way to use $compile, and making a directive called <loading-splash />

**I'm ok with this being written like this way,** since we really only use the loading-splash in one place, but all the jquery DOM manipulation kinda bugs me....

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,274 @@
...
+            }
+            open = false;
+        };
+
+        this.isOpen = function () {
+            return open;
+        };
+    });
+
+    uiServices.factory('ModalFactory', function (LoadingModal, BootstrapDialogManager) {
```
nickdotreid: The ModalFactory shouldn't rely on the LoadingModal... lets keep them separate...

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -2,9 +2,274 @@
...
+        };
+
+        modalFactory.angularHandler = function(title, defaultMsg, callback) {
+            return function(response) {
+                return modalFactory.handleResponse(title, defaultMsg, response, callback);
+            };
+        };
+
+        modalFactory.handleWithStackTrace = function(title, defaultMsg, response) {
+            var msg = "server.error";
```
nickdotreid: Isn't the msg variable actually the defaultMsg, and the defaultMsg parameter the msg argument (which motechAlert is the first argument)

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3629,17 +3629,17 @@
                     $http.get('../mds/instances/' + relatedEntityId + '/instance/' + instanceId).success(function (data) {
                         $scope.editRelatedFields = angular.copy(data.fields);
                     }).error(function(response) {
-                        handleResponse('mds.error', 'mds.error.instancesList', response);
+                        ModalFactory.handleResponse('mds.error', 'mds.error.instancesList', response);
                     });
                 },
                 setExistingWithoutId = function () {
```
nickdotreid: How is this setExistingWithoutId function working within the context of the callback?

nickdotreid:
@wstrzelczyk @sebbrudzinski 

I made a couple of comments in the code... please feel free to push back on any of it -- I feel that I'm most worried about keeping the exposed functions as clear as possible for someone else using it...

Functions, like the handler functions, still throw me because I'm not sure what they do (just by reading the function name). I think it would be more clear to wrap the error handling in place, and expose that these functions show an Alert (or ModalFactory.motechAlert)

I also don't get the difference between ModalFactory.motechAlert and ModalFactory.alert -- I can tell that ModalFactory.alert takes an object and that it is used to change the Modal type in some places, but that seems to be the largest difference (please correct me if I'm wrong)

IMHO: motechAlert() and alert() should be the same function... and this argument carries to motechConfirm()

I like how ModalFactory.errorAlert() is different in that it changes the Modal type in a single place, and what's different is obvious in the name. I could see a warningAlert function working in the same way...

 This last request is a stretch, and if you can ignore it I won't mind -- alert() and confirm() should return a promise ($q), instead of a callback function... I feel this makes functions easier to read because new functionality isn't mixed into parameter definitions.

sebbrudzinski:
Conflicts must be resolved.

sebbrudzinski:
@nickdotreid Waldek has addressed your comments in his commits. Can you check and see if there are still any problems?

nickdotreid:
@sebbrudzinski @wstrzelczyk 

I'm happy with how ModalFactory is constructed, and the places I that use it seem right....

I'd nitpick that we don't need 'show' prefixed on all the ModalFactory method calls .... but it makes sense because the method does "show" a modal... so I'm happy to leave it be

There are a few spots in the code that use ModalFactory.alert() ... I'm assuming those still need to be updated, right?

wstrzelczyk:
No, I updated all.

sebbrudzinski:
@wstrzelczyk I can see 5 occurences of `ModalFactory.alert` in the repository.. Can you check and fix?

PR #493, opened 2016-04-13
MOTECH-2094: Added a migration to scheduler module

kmadej:
After scheduler refactor, class service.impl.MotechScheduledJob changed package
to service.MotechScheduledJob. This class name is stored in motechquartz database,
so we lost backward compatibility. This migration should fix the problem and now
old jobs can be properly fired after scheduler API clean up.


modules/scheduler/scheduler/src/main/resources/db/migration/scheduler/default/V3__MOTECH-2094.sql
```diff
@@ -0,0 +1,2 @@
+
```
sebbrudzinski: Some comment on why we do that would be in place here.

modules/scheduler/scheduler/src/main/resources/db/migration/scheduler/mysql/V3__MOTECH-2094.sql
```diff
@@ -0,0 +1,2 @@
+
```
sebbrudzinski: Some comment on why we do that would be in place here.

sebbrudzinski:
test this psql

sebbrudzinski:
Doesn't seem to work with PostgreSQL

kmadej:
test this psql

PR #506, opened 2016-04-20
MOTECH-2390: Fixed delete option for permissions in WebSecurity

jredlarski:
Added delete option for all permissions in Security (permission.html, websecurity.css). If permission is default delete option is disabled and an appropriate message is shown when 'delete' is hovered.
Added content for the message in messages.properties.


platform/web-security/src/main/resources/webapp/messages/messages.properties
```diff
@@ -115,6 +115,7 @@ security.confirmationRequired=Password confirmation is required
 security.emailTaken=Email already in use
 security.confirm=Confirm
 security.confirm.permissionDelete=Delete this permission?
+security.defaultPermission=You can not delete a default permission
```
pgesek: maybe change this to 'predefined' permission - default doesn't really fit that well here

PR #507, opened 2016-04-21
MOTECH-2389: Fixed more than one same warning in Recent Task Activity

weronika-ciecierska:
Warning about null object of given type is shown only once in recent task activity.
Message of warning specifies object of what type cannot be found.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/util/DataSourceObject.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/util/DataSourceObject.java
@@ -38,9 +38,13 @@ public class DataSourceObject {
         return failIfNotFound;
     }
 
-    public boolean isNullWarningPublished() { return nullWarningPublished; }
+    public boolean isNullWarningPublished() {
+        return nullWarningPublished;
+    }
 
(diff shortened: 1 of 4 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/util/DataSourceObject.java
```diff
@@ -36,6 +38,10 @@ public boolean isFailIfNotFound() {
         return failIfNotFound;
     }
 
+    public boolean isNullWarningPublished() { return nullWarningPublished; }
```
sebbrudzinski: Why is this condensed to one line? The statement should start in a new line, similar to the closing brace.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/util/DataSourceObject.java
```diff
@@ -36,6 +38,10 @@ public boolean isFailIfNotFound() {
...
+
+    public void setNullWarningPublished(boolean nullWarningPublished) { this.nullWarningPublished = nullWarningPublished; }
```
sebbrudzinski: Why is this condensed to one line? The statement should start in a new line, similar to the closing brace.

PR #509, opened 2016-04-25
MOTECH-2395: Serving NullPointerException in Lookup

1stmateusz:


platform/mds/mds/src/main/java/org/motechproject/mds/domain/Lookup.java
```diff
@@ -155,9 +160,13 @@ public LookupDto toDto() {
...
             boolean useGenericParam = toBoolean(getUseGenericParams().get(lookupFieldName));
-            LookupFieldDto lookupField = new  LookupFieldDto(field.getId(), field.getName(), lookupFieldType, customOperator,
-                    useGenericParam, LookupName.getRelatedFieldName(lookupFieldName));
-            lookupFields.add(lookupField);
+            if (field != null) {
+                LookupFieldDto lookupField = new LookupFieldDto(field.getId(), field.getName(), lookupFieldType, customOperator,
+                        useGenericParam, LookupName.getRelatedFieldName(lookupFieldName));
+                lookupFields.add(lookupField);
+            } else {
+                LOGGER.info("Can't create LookupFieldDto from field with name {}", lookupFieldName);
```
pgesek: I believe we should rather from an exception in this case - with the correct message. Proceeding should not be possible in this case.

platform/mds/mds/src/main/java/org/motechproject/mds/domain/Lookup.java
```diff
@@ -268,6 +277,7 @@ public final Field getLookupFieldByName(String name) {
                 return field;
             }
         }
+        LOGGER.info("Can't get field with name {}", name);
```
pgesek: this should be a warn

PR #510, opened 2016-04-25
MOTECH-2348: comparing files path without URL conversion

1stmateusz:
It seams that URL encoding on Jenkins sometimes fails, so test no more uses
URL


1stmateusz:
test this please

1stmateusz:
test this please

PR #511, opened 2016-04-25
Fixed motech build

kmadej:
This is temporary solution. There is some problem with jquery-cookie library.
If you really need a quick solution, please use my changes.


kmadej:
I closed PR, because the issue has been resolved.

PR #520, opened 2016-04-28
MOTECH-2363 Change TaskDataProviderValidator doesnt throw ValidationE…

lkukielka:
…xception for empty provider

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImpl.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImpl.java
@@ -72,15 +72,19 @@ public class TaskDataProviderServiceImpl implements TaskDataProviderService, Osg
         final Type type = new TypeToken<TaskDataProvider>() { } .getType();
         final TaskDataProvider provider = (TaskDataProvider) motechJsonReader.readFromStream(stream, type);
 
-        LOGGER.info("Registering a task data provider with name: {}", provider.getName());
+        if(TaskDataProviderValidator.validateIsNotEmpty(provider)) {
+            LOGGER.info("Registering a task data provider with name: {}", provider.getName());
 
-        Set<TaskError> errors = TaskDataProviderValidator.validate(provider);
+            Set<TaskError> errors = TaskDataProviderValidator.validate(provider);
 
-        if (!isEmpty(errors)) {
-            throw new ValidationException(TaskDataProviderValidator.TASK_DATA_PROVIDER, errors);
+            if (!isEmpty(errors)) {
+                throw new ValidationException(TaskDataProviderValidator.TASK_DATA_PROVIDER, errors);
+            }
+
+            addProvider(provider);
+        } else {
+            LOGGER.info("Registering an empty task data provider with name: {} is not possible", provider.getName());
         }
-
-        addProvider(provider);
     }
 
     @Override
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskDataProviderValidator.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskDataProviderValidator.java
@@ -33,7 +33,7 @@ public final class TaskDataProviderValidator extends GeneralValidator {
      * @return  the set of encountered errors
      */
     public static boolean validateIsNotEmpty(TaskDataProvider provider) {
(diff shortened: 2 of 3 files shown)


modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImplTest.java
```diff
@@ -63,7 +63,7 @@ public void setup() throws Exception {
         ((TaskDataProviderServiceImpl) taskDataProviderService).bind(dataProviderDataService, Collections.emptyMap());
     }
 
-    @Test(expected = ValidationException.class)
+    @Test(expected = WantedButNotInvoked.class)
```
sebbrudzinski: Why are we expecting that the test throws exception..?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskDataProviderValidator.java
```diff
@@ -32,17 +32,17 @@ private TaskDataProviderValidator() {
      * @param provider  the data provider for validation, not null
      * @return  the set of encountered errors
      */
+    public static boolean validateIsNotEmpty(TaskDataProvider provider) {
+        return !checkEmpty(new HashSet<>(), TASK_DATA_PROVIDER, "objects", provider.getObjects());
```
sebbrudzinski: It doesn't seem to make sense to expose this method. After all, we only check if provider.getObjects() is empty and return either true or false. We don't use generated errors later on, so this can be completely removed.

lkukielka:
test this please

modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskDataProviderValidator.java
```diff
@@ -32,17 +32,17 @@ private TaskDataProviderValidator() {
      * @param provider  the data provider for validation, not null
      * @return  the set of encountered errors
      */
+    public static boolean validateIsNotEmpty(TaskDataProvider provider) {
```
sebbrudzinski: javadoc is incorrect

modules/tasks/tasks/src/main/java/org/motechproject/tasks/validation/TaskDataProviderValidator.java
```diff
@@ -32,17 +32,17 @@ private TaskDataProviderValidator() {
...
+        return !provider.getObjects().isEmpty();
+    }
+
     public static Set<TaskError> validate(TaskDataProvider provider) {
```
sebbrudzinski: lacks javadoc

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImpl.java
```diff
@@ -72,15 +72,19 @@ public void registerProvider(final InputStream stream) {
         final Type type = new TypeToken<TaskDataProvider>() { } .getType();
         final TaskDataProvider provider = (TaskDataProvider) motechJsonReader.readFromStream(stream, type);
 
-        LOGGER.info("Registering a task data provider with name: {}", provider.getName());
+        if(TaskDataProviderValidator.validateIsNotEmpty(provider)) {
```
sebbrudzinski: We should have a test that checks the registration of empty provider

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImplTest.java
```diff
@@ -6,6 +6,7 @@
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
+import org.mockito.exceptions.verification.WantedButNotInvoked;
```
sebbrudzinski: Unused

modules/tasks/tasks/src/test/java/org/motechproject/tasks/service/impl/TaskDataProviderServiceImplTest.java
```diff
@@ -16,6 +17,7 @@
 import org.motechproject.tasks.exception.ValidationException;
 import org.motechproject.tasks.repository.DataProviderDataService;
 import org.motechproject.tasks.service.TaskDataProviderService;
+import org.powermock.api.mockito.PowerMockito;
```
sebbrudzinski: Unused(?)

sebbrudzinski:
@pgesek LGTM

PR #528, opened 2016-04-30
MOTECH-1893 Create additional IT tests for cascade delete

atish160384:
IT tests for behavior of cascade delete for Uni Directional, Bidirectional, Map Relationship.


pgesek:
@sebbrudzinski can you review this?

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -500,6 +543,90 @@ public void testCascadeDelete() {
...
+        assertNotNull(mother);
+
+        ChildCase child = childCaseDataService.findById(child1.getId());
+        assertNotNull(child);
+
+        motherCaseDataService.delete(mother);
+        mother = motherCaseDataService.findById(mother.getId());
+        assertNull(mother);
+
+        child = childCaseDataService.findById(child2.getId());
```
sebbrudzinski: Why don't we also verify that another child was deleted?

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -500,6 +543,90 @@ public void testCascadeDelete() {
     }
 
     @Test
+    public void testCascadeDeleteUniDirectional() {
+        ChildCase child1 = new ChildCase("Child 1");
+        childCaseDataService.create(child1);
+
+        ChildCase child2 = new ChildCase("Child 2");
+        childCaseDataService.create(child2);
```
sebbrudzinski: Side note - it shouldn't be required to manually create related objects. They should be by default created with the owning instance (mother).

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -500,6 +543,90 @@ public void testCascadeDelete() {
...
+
+        event = eventDataService.findById(event.getId());
+        assertNotNull(event);
+
+        eventLogDataService.delete(eventLog);
+        eventLog = eventLogDataService.findById(eventLog.getId());
+        assertNull(eventLog);
+
+        event = eventDataService.findById(event.getId());
+        assertNull(event);
```
sebbrudzinski: This doesn't seem to check the map at all, even though the test name says so. It checks cascade delete between EventLog and Event, which is unidirectional relationship.

platform/mds/mds-test-bundle/src/test/java/org/motechproject/mds/test/osgi/MdsDdeBundleIT.java
```diff
@@ -500,6 +543,90 @@ public void testCascadeDelete() {
...
+
+        Unit unit = new Unit("Tree");
```
sebbrudzinski: Seems like all of the below belongs to a new test. Also not sure why are we testing unidirectional relationships again.

sebbrudzinski:
Besides the inline comments, I'm still not sure if we already test all cases. Can we refactor test names for cascade deletes to describe the test cases? We should test one-to-one, one-to-many (uni- and bi- directional) and many-to-many relationships and of course maps.

PR #536, opened 2016-05-07
MOTECH-2296 Task retries for event-based actions

sebbrudzinski:
This commit enhances the task retries for the event-based
task actions. Default event system retries are no longer
invoked for task actions. The event module allows to register
event callbacks now, that will be invoked when the event
is handled. Moreover, the Task module now keeps the track
of the task executions.


modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/TaskActivity.java
```diff
@@ -61,7 +67,7 @@ public TaskActivity() {
      * @param activityType  the activity type
      */
     public TaskActivity(String message, Long task, TaskActivityType activityType) {
-        this(message, new ArrayList<String>(), task, activityType);
+        this(message, new ArrayList<>(), task, activityType, 0);
```
pgesek: a task can't have 0 actions if I recall correctly

modules/tasks/tasks/src/main/java/org/motechproject/tasks/constants/EventDataKeys.java
```diff
@@ -29,4 +29,7 @@ private EventDataKeys() {
     public static final String REPEAT_INTERVAL_TIME = "repeatIntervalInSeconds";
     public static final String JOB_SUBJECT = "jobSubject";
     public static final String TASK_ID = "task_ID";
+    public static final String TASK_ACTIVITY_ID = "task_activity_ID";
```
pgesek: if these are included in every task action event, we should prefix them at least. I am weary of adding parameters to every event because of the issues we already had in DHIS due to retry metadata being added as task params, we will probably have the same issue with these params sooner or later. Is there a way to avoid this?

pgesek: At the worst we can consider extending MotechEvent with a metadata map and put params like these there

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -30,64 +37,120 @@ public TaskActivityServiceImpl(TaskActivitiesDataService taskActivitiesDataServi
...
 
     @Override
-    public void addError(Task task, TaskHandlerException e, Map<String, Object> parameters) {
-        taskActivitiesDataService.create(new TaskActivity(e.getMessage(), e.getArgs(), task.getId(),
-                TaskActivityType.ERROR, ExceptionUtils.getStackTrace(e), parameters));
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public long addTaskStarted(Task task, Map<String, Object> parameters) {
+        int totalActions = task.getActions().size();
+        TaskActivity activity = taskActivitiesDataService.create(
+                new TaskActivity(TASK_IN_PROGRESS, Arrays.asList("0", String.valueOf(totalActions)), task.getId(), TaskActivityType.IN_PROGRESS, null, parameters, totalActions));
```
pgesek: what are saving here exactly with the "0" and totalActions count?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -30,64 +37,120 @@ public TaskActivityServiceImpl(TaskActivitiesDataService taskActivitiesDataServi
     }
 
     @Override
-    public void addError(Task task, TaskHandlerException e, Map<String, Object> parameters) {
-        taskActivitiesDataService.create(new TaskActivity(e.getMessage(), e.getArgs(), task.getId(),
-                TaskActivityType.ERROR, ExceptionUtils.getStackTrace(e), parameters));
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
pgesek: why is this done outside of the existing transaction?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -30,64 +37,120 @@ public TaskActivityServiceImpl(TaskActivitiesDataService taskActivitiesDataServi
...
+        }
+
+        TaskExecutionProgress progress = activity.getTaskExecutionProgress();
+        progress.addSuccess();
+        boolean taskFinished = progress.getActionsSucceeded() == progress.getTotalActions();
+
+        if (taskFinished) {
+            activity.setActivityType(TaskActivityType.SUCCESS);
+            activity.setMessage(TASK_SUCCEEDED);
+            activity.getFields().clear();
```
pgesek: why are we clearing fields for the activity?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/constants/EventDataKeys.java
sebbrudzinski: I don't really think there's any way to avoid this. MotechEvent is basically everything that a MOTECH instance receives so that metadata must live somewhere in the MotechEvent, so we know which task/activity the event is connected to. Metadata map sounds like a good plan, so I can proceed with that.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
pgesek: Is this because of a retry? Do we just keep updating the same activity? I would imagine I would have  a list of all failures, same as we have with regular retry, or am I missing something?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
sebbrudzinski: We need to have the activity commited, so it can be used later on to update the status of the task execution.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
sebbrudzinski: This is used to generate the message on the view, about the number of successfully executed actions.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
pgesek: So the problem is that retries start before the task transaction finishes?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
pgesek: ok, I get it, this something like 0/5 type of thing

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
sebbrudzinski: Not necessarily retries, but just regular task executions.

pgesek: I don't get it, why would other executions need to read this data from the db? Do tasks need to check activities now when triggering?

What are the consequences of doing this in a different tx? Can something go corrupt? I'm thing the failure counter in the task and the activity list, but is there something else?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/constants/EventDataKeys.java
pgesek: I think we can do the metadata map for scheduler/tasks params, so that they don't clutter the regular params.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
sebbrudzinski: No. What Tasks module needs to do after successful or failed execution is to update the ->existing<- activity. Since the task actions can be either events or service calls, it might happen that some actions (event-based) are executed even before the Tasks module is done starting other actions. In that case, the executed actions would fail to update their respective activity, since they are not commited in the database yet. Creating initial task activity is done right before executing actions (in the same transaction, if we do not start new one here).

modules/tasks/tasks/src/main/java/org/motechproject/tasks/constants/EventDataKeys.java
pgesek: Can we split this event stuff from this PR though? This is too big already

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
sebbrudzinski: What do you mean with "same as with regular retry"? This is not connected to the retries. This handles the execution of (several) actions within one task. And currently we keep only one failure per task. Should we adjust this as a part of this ticket? This basically clears the previous message "Executed X of Y actions" and replaces it with "All actions have been executed successfuly".

sebbrudzinski:
A fix for a subticket must be merged first, so we need to hold this off for now.

sebbrudzinski:
@PiotrZurawskiSolDevelo  This pull request requires a thorough verification of the Tasks module

PiotrZurawskiSolDevelo:
I checked this PR. I Created few scenarios with tasks:
What works well:
1. I created Task with few data sources, few filters and few actions,
2. I created Task with 7 actions. When I run it, and I go to tasks, in Recent Task appear new Icon. That inform us about that tasks is not completely done. When We click on this icon, in history is appear information how many actions was done and how many left.

What is not working well.
1. I created task with 6 actions : Send Email. In task I check checkbox: Retry task on failure, and I set Retry Count : 3 and Retry Interval : 40. 
**When I run task, in Recent task activity I can see all failures for all actions,** here should be visible only one task activity for one task.
2. When I run task I can see in email logs 42 records ( I had clean database).
3. I had created task with 2 actions send email. In Send Email fields I had used bubbles from trigger.
When task run, in Email logs I can see data from bubbles only in action from first run. 
4. When task have set checkbox for Retry task on failure and Retry Count = 3, Retry interval = 40 and in task is created 7 actions, I can not see information about that how many actions was done and how many left in this task (but it was done for all actions).

sebbrudzinski:
4. This is intended. When the task fails, the info about number of task executions is replaced with the exception that has caused the failure.

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActionExecutor.java
```diff
@@ -138,6 +138,9 @@ private ActionEvent getActionEvent(TaskActionInformation actionInformation)
                         try {
                             String userInput = keyEvaluator.evaluateTemplateString(template);
                             Object obj = actionParameter.getType().parse(userInput);
+                            if (obj.equals("") && keyEvaluator.getTaskContext().getTriggerParameters().containsKey(key)) {
```
sebbrudzinski: Let's use StringUtils.isBlank()

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActionExecutor.java
```diff
@@ -138,6 +138,9 @@ private ActionEvent getActionEvent(TaskActionInformation actionInformation)
...
+                                obj = keyEvaluator.getTaskContext().getTriggerParameters().get(key);
+                            }
```
sebbrudzinski: What about LIST and MAP cases? It seems like they won't get included as well.

sebbrudzinski:
@PiotrZurawskiSolDevelo Please test this with new changes

PiotrZurawskiSolDevelo:
I have checked:
1. It's working now.
2. It's working now.
3. Still it's not working. I can not see data when task is retried.

mkruszynski:
@PiotrZurawskiSolDevelo Can you give used bubbles?

PiotrZurawskiSolDevelo:
@mkruszynski I used all bubbles from trigger (Create Motech Permission).

sebbrudzinski:
@PiotrZurawskiSolDevelo The fixes have been uploaded

PiotrZurawskiSolDevelo:
@sebbrudzinski I checked this and now it looks like everything is working well.

PR #538, opened 2016-05-09
MOTECH-1600 fixing two errors in the install doc.

craigappl:
The install doc had an error with the restructured text ..note annotation and an error that caused the curl to fail when downloading the .war file from nexus.


craigappl:
FYI @llavoie  can you code review this?

PR #543, opened 2016-05-13
MOTECH-2435 Changed imports for @JsonIgnore annotation

lkukielka:


sebbrudzinski:
I'm putting this on hold for now. This might be an issue with all JsonIgnore annotation in the project, so we might need to tackle the problem in more general way.

sebbrudzinski:
The problem was actually with DHIS2 module using some fasterxml imports (which made codehaus jackson stop working in Tasks, when DHIS2 was started). DHIS2 now uses only codehaus Jackson, like all of MOTECH what resolved the problem.

PR #545, opened 2016-05-13
MOTECH-2484 Changing language to 'tw.big5' breaks MOTECH - hide the C…

wstrzelczyk:
…hinese option

### DIFF
--- a/platform/server-bundle/src/main/resources/webapp/js/controllers.js
+++ b/platform/server-bundle/src/main/resources/webapp/js/controllers.js
@@ -396,7 +396,10 @@
                 $scope.i18n = data;
             }),
             $scope.doAJAXHttpRequest('GET', 'lang/list', function (data) {
-                if(data['zh_TW.Big5'] !== undefined) { delete data['zh_TW.Big5']; }
+                // "TODO: Temporarily hiding the Chinese option, until it's fixed with MOTECH-2484"
+                if(data['zh_TW.Big5'] !== undefined) {
+                    delete data['zh_TW.Big5'];
+                }
                 $scope.languages = data;
             }),
             $scope.doAJAXHttpRequest('GET', 'lang', function (data) {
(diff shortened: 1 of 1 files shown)


wstrzelczyk:
Added hiding the Chinese option

platform/server-bundle/src/main/resources/webapp/js/controllers.js
```diff
@@ -396,6 +396,7 @@
                 $scope.i18n = data;
             }),
             $scope.doAJAXHttpRequest('GET', 'lang/list', function (data) {
+                if(data['zh_TW.Big5'] !== undefined) { delete data['zh_TW.Big5']; }
```
sebbrudzinski: Let's style this code properly. The content of the if, as well as the closing brace should be in a new line. Moreover, let's add a comment with the ticket number, explaining that we temporarily hide this, until it's fixed.

platform/server-bundle/src/main/resources/webapp/js/controllers.js
```diff
@@ -414,6 +415,7 @@
                 $scope.doAJAXHttpRequest('GET', 'lang/locate', function (data) {
                     $scope.i18n = data;
                 }), $scope.doAJAXHttpRequest('GET', 'lang/list', function (data) {
+                    if(data['zh_TW.Big5'] !== undefined) { delete data['zh_TW.Big5']; }
```
sebbrudzinski: Let's style this code properly. The content of the if, as well as the closing brace should be in a new line. Moreover, let's add a comment with the ticket number, explaining that we temporarily hide this, until it's fixed.

platform/server-bundle/src/main/resources/webapp/js/controllers.js
```diff
@@ -528,6 +530,11 @@
 
         $scope.getStartupViewData = function() {
             $scope.doAJAXHttpRequest('GET', '../server/startupviewdata', function (data) {
+                if(data.languages['zh_TW.Big5'] !== undefined) { delete data.languages['zh_TW.Big5']; }
```
sebbrudzinski: Let's style this code properly. The content of the if, as well as the closing brace should be in a new line. Moreover, let's add a comment with the ticket number, explaining that we temporarily hide this, until it's fixed.

PR #546, opened 2016-05-13
MOTECH-2054 - Rollback changes connected with time server and server …

wstrzelczyk:
…up time


sebbrudzinski:
@pgesek OK to merge

PR #548, opened 2016-05-17
MOTECH-2494 Name incorrect display in DataServices

wstrzelczyk:


pgesek:
We should not be hardcoding this in mds-web. There are annotations like UIDisplayable that can be used to hide a field.

PR #550, opened 2016-05-19
MOTECH-2449: fix Date Time problem in footer

1stmateusz:


kmadej:
@1stmateusz can you wrtie a brief comment why this change helps ?

1stmateusz:
There was a problem with response after getime request. Without any modules installed server responded with only milliseconds, but with all modules installed it responded with JSON object that contained additional info about data etc, but we actually used only milliseconds value. I've changed server response and now it always sends only milliseconds value and time value is properly displayed with and without modules installed.

kmadej:
we should find a real problem here, why server returns different values depends on installes modules ?

sebbrudzinski:
I agree with Kamil here. A similar situation took place in the DHIS2 module, where installing it would make Jackson serializer to stop working in Tasks. Perhaps, the Metrics module also affects the Jackson serialization process somehow. The serialization should work the same way, no matter what modules are added.

kmadej:
The changes for this bug are now available in https://github.com/motech/modules/pull/294

PR #552, opened 2016-05-19
MOTECH-2523 Fixed importing csv for mds entity

lkukielka:


craigappl:
Performed QA this morning and the import succeeded. Thanks for the quick turn-around!

PR #556, opened 2016-05-24
MOTECH-2479 Unintuitive error when Lookup return more than one result

wstrzelczyk:


platform/mds/mds/src/main/java/org/motechproject/mds/lookup/LookupExecutor.java
```diff
@@ -61,8 +61,10 @@ public Object execute(Map<String, ?> lookupMap, QueryParams queryParams) {
             return MethodUtils.invokeMethod(dataService, lookup.getMethodName(),
                     args.toArray(new Object[args.size()]),
                     argTypes.toArray(new Class[argTypes.size()]));
-        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+        } catch (NoSuchMethodException | IllegalAccessException e) {
             throw new LookupExecutorException("Unable to execute lookup " + lookup.getLookupName(), e);
+        } catch (InvocationTargetException e) {
+            throw new LookupExecutorException("Unable to execute lookup " + lookup.getLookupName() + ". " + e.getTargetException().getMessage() + ".", e);
```
sebbrudzinski: We cannot just retrieve the not localizable message from the underlying exception and display it to the user. We must verify ourselves that the problem is connected to the lookup execution and wrap it into our own, localizable exception, that clearly explains the problem.

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3276,7 +3276,13 @@
             exportWithOrder : false
         };
 
-        $scope.setDataRetrievalError = function (value) {
+        $scope.setDataRetrievalError = function (value, responseText) {
+            if(responseText) {
+                $scope.retrievalErrorText = $scope.msg(responseText.replace('key:', '').trim());
+            } else {
+               $scope.retrievalErrorText = $scope.msg('mds.error.lookupExecError');
```
sebbrudzinski: Why do we hardcode lookup execution error in case no response text is provided. Shouldn't the error/message always come from the server?

wstrzelczyk:
@sebbrudzinski We do not need this part of code.

platform/mds/mds/src/main/java/org/motechproject/mds/lookup/LookupExecutor.java
```diff
@@ -61,8 +66,22 @@ public Object execute(Map<String, ?> lookupMap, QueryParams queryParams) {
...
+                lookupExceptionMessageKey = "mds.error.lookupExecUserError";
+                for (Throwable exception : userException.getNestedExceptions()) {
+                    if (exception instanceof QueryNotUniqueException) {
+                        lookupExceptionMessageKey = "mds.error.lookupExecNotUniqueError";
+                        throw new LookupExecutorException(lookupExceptionMessage + lookup.getLookupName() + ".", e, lookupExceptionMessageKey);
+                    }
+                }
+                throw new LookupExecutorException(lookupExceptionMessage + lookup.getLookupName() + ".", e, lookupExceptionMessageKey);
+            } else {
+                throw new LookupExecutorException(lookupExceptionMessage + lookup.getLookupName() + ".", e, lookupExceptionMessageKey);
```
sebbrudzinski: From what I can see, all these throws look exactly the same. Only the message differs. Let's move all of those throws outside the if blocks and only set the message keys there.

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/service/impl/InstanceServiceImpl.java
```diff
@@ -306,7 +314,7 @@ public long countRecordsByLookup(Long entityId, String lookupName, Map<String, O
             LookupExecutor lookupExecutor = new LookupExecutor(service, lookup, fieldMap);
             return lookupExecutor.executeCount(lookupMap);
         } catch (RuntimeException e) {
-            throw new LookupExecutionException(e);
+            throw new LookupExecutionException(e, "mds.error.lookupExecError");
```
sebbrudzinski: This should also use a constant

PR #558, opened 2016-05-25
MOTECH-2525: Fixed: Random web-security failure in spring 3.2

pkornowski:
I think, the problem was with schemaLocation. If a version is not specified before suffix then we gets the latest version (actually 4.2).


kmadej:
can you add a version also to spring-cache and spring-util

PR #559, opened 2016-05-25
MOTECH-2296 Ability to register callbacks to MotechEvents

sebbrudzinski:
This commit adds the ability to register callback services, that
will be invoked when the execution of a MotechEvent is finished.
The service must implement the EventCallbackService and provide
a callback name, that must be also included in the MotechEvent.
The interface provides two methods - one for success callbacks
and another one for failure callbacks.

### DIFF
--- a/platform/event/src/main/java/org/motechproject/event/listener/impl/ServerEventRelay.java
+++ b/platform/event/src/main/java/org/motechproject/event/listener/impl/ServerEventRelay.java
@@ -1,12 +1,18 @@
 package org.motechproject.event.listener.impl;
 
+import org.apache.commons.lang.StringUtils;
 import org.motechproject.event.MotechEvent;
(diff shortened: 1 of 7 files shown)


platform/event/src/main/java/org/motechproject/event/listener/impl/ServerEventRelay.java
```diff
@@ -146,31 +155,65 @@ public void handleEvent(Event osgiEvent) {
...
+
+            for (ServiceReference<EventCallbackService> ref : references) {
+                EventCallbackService callback = bundleContext.getService(ref);
+                if (callback.getName().equals(callbackName)) {
+                    return callback;
+                }
+            }
+        } catch (InvalidSyntaxException e) {
+            //Should never happen
+            LOGGER.error("Passed filter expression is incorrect.");
```
pgesek: Unimportant, but also log e

PR #562, opened 2016-05-25
fix entering module settings in admin module

1stmateusz:
After adding state provider there was an url parsing error and angular couldn't find proper state


kmadej:
Can you change the name of PR ? This fix is not connected with MOTECH-2461

platform/server-bundle/src/main/resources/webapp/js/controllers.js
```diff
@@ -222,6 +222,7 @@
             var convertUrl = function (urlParam) {
                 if(urlParam.indexOf('/') === 0) {urlParam = urlParam.replace('/', '');}
                 if(urlParam.indexOf('/') > 0) {urlParam = urlParam.replace('/', '.');}
+                urlParam = urlParam.split('/')[0];
```
kmadej: can you use remove function, I think it will be more readable

PR #566, opened 2016-05-30
MOTECH-2543: Fixed problem with key:value map in the Tasks module

ngraczewski:


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActionExecutor.java
```diff
@@ -162,18 +162,15 @@ private ActionEvent getActionEvent(TaskActionInformation actionInformation)
...
             switch (array.length) {
                 case 2:
                     array[1] = array[1].trim();
-                    mapKey = getValue(array[0]);
-                    mapValue = getValue(array[1]);
 
                     tempMap.put(
-                        ParameterType.getType(mapKey.getClass()).parse(keyEvaluator.evaluateTemplateString(array[0])),
-                        ParameterType.getType(mapValue.getClass()).parse(keyEvaluator.evaluateTemplateString(array[1]))
+                        keyEvaluator.evaluateTemplateString(array[0]),
```
pgesek: we should add a unit test for the case that this fixes

PR #573, opened 2016-06-02
Added new migration to fixed module migrations feature.

pkornowski:
At first I added auto-increment in base version, but it should be placed in a separate migration. In PSQL I used sequence. I looked for solution with adding 'serial' to existing field in table, like in MySQL, but I have found only this.

### DIFF
--- a/platform/mds/mds/src/main/resources/db/migration/default/V6__MOTECH-2578.sql
+++ b/platform/mds/mds/src/main/resources/db/migration/default/V6__MOTECH-2578.sql
@@ -1,4 +1,4 @@
--- Added auto increment sequence
+-- Changed primary key to serial type.
 
 ALTER TABLE "MigrationMapping" DROP COLUMN "flywayMigrationVersion";
 ALTER TABLE "MigrationMapping" ADD COLUMN "flywayMigrationVersion" serial NOT NULL;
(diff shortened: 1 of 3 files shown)


platform/mds/mds/src/main/resources/db/migration/default/V6__MOTECH-2578.sql
```diff
@@ -0,0 +1,22 @@
+-- Added auto increment sequence
+
```
pgesek: Isn't this a bit overcomplicated? Can't we just use the serial type?

kmadej:
test this psql

platform/mds/mds/src/main/resources/db/migration/default/V6__MOTECH-2578.sql
pkornowski: @pgesek 
I looked for solution with adding serial type. I found some solution which works, when we add a new field in the table, but when we want to change type in an existing field, it doesn't work.

pkornowski:
test this psql

platform/mds/mds/src/main/resources/db/migration/default/V6__MOTECH-2578.sql
```diff
@@ -0,0 +1,5 @@
+-- Added auto increment sequence
```
kmadej: Changed primary key to serial type

pkornowski:
@pgesek 
Can we merge this?

kmadej:
@pgesek 
Looks good to me, can we merge this ?

PR #577, opened 2016-06-07
Catch loss of authenticated session in UI

nickdotreid:
I've started adding code to catch the loss of an authenticated session in MOTECH.

This solution works for Ajax queries, but I have not figured out how to make it work when resources are loaded -- as in the case of loading a new module with ocLazyLoad

The new MOTECH-UI uses a different process to authenticate a user, so this PR is targeting the 'legacy' UI and hopes to provide a better experience than the page simply breaking.

I would love opinions about:
(A) MOTECH should show an error message when a user's session is lost
(B) Where we should insert code in the module loading process to catch failed (redirected) resources
 -- 1) Before loading a module, ping getUser to verify session (easy, but makes code less specific)
 -- 2) See if ocLazyLoad promise knows the redirect happened (not sure this will work)
 -- 3) Catch resource redirected event in browser (not sure how to do this)
 -- 4) Change permissions for resource files so unauthenticated users can load them (I like this)
 -----> (4) will have effect of allowing user to browse around unauthenticated until any data is loaded from server side (which would happen immediately in most modules (not the email module))

@pgesek @wstrzelczyk @sebbrudzinski @jredlarski @llavoie @koshalt @frankhuster (or well anyone else)

### DIFF
--- a/ui/src/js/auth-interceptor.js
+++ b/ui/src/js/auth-interceptor.js
@@ -1,0 +1,14 @@
+(function(open) {
+/*
+ * Catches all XMLHttpReqests, and if 302 to /login, sets window location to that page
+ */
+  XMLHttpRequest.prototype.open = function(method, url, async, user, pass) {
+    var xhr = this;
+    this.onload = function(){
+        if(window.location.pathname.indexOf("server/login") < 0 && (xhr.status === 302 || xhr.status === 401)){
+        	document.location.reload();
+        }
+    }
+    open.call(this, method, url, async, user, pass);
+  };
+})(XMLHttpRequest.prototype.open);


pgesek:
@nickdotreid Ideally it should only ask for the username/password again when the session expires. 
Regarding question B, (4) sounds fine to me, I would just choose the easiest route.

pgesek:
@nickdotreid I had to revert this, because it made Motech unusable and the UI would constantly refresh. Please reopen this. Thanks.

nickdotreid:
haha, will do — I'll test it further and see what's going on...

On Wed, Jun 15, 2016 at 5:57 AM, Paweł Gesek <notifications@github.com>
wrote:

> @nickdotreid <https://github.com/nickdotreid> I had to revert this,
> because it made Motech unusable and the UI would constantly refresh. Please
> reopen this. Thanks.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/motech/motech/pull/577#issuecomment-226178943>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAoU-y_9hrEm-HqszLbF50Z6dZZ01yhAks5qL_a8gaJpZM4IwVwn>
> .
>



-- 

Nick Reid
Grameen Foundation, Mobile Health Innovations
2101 4th Ave | Suite 1550 | Seattle, WA 98121
nreid@grameenfoundation | +1.510.410.0020 | Skype: nickdotreid
GrameenFoundation.org <http://grameenfoundation.org/> | Twitter
<http://twitter.com/grameenfdn>│MOTECH Project
<http://www.motechproject.org/>
*Connecting the World's Poor to Their Potential*

PR #357, opened 2016-02-21
MOTECH-2020 Refactored task field bubbles

nickdotreid:
This is an initial commit to get the code review process going. There are minor bugs when interacting with the field bubble popover, and large chunks of now legacy code that will be removed.

PR #579, opened 2016-06-08
MOTECH-2542: fix string manipulations in task module for task filters

1stmateusz:
After task [field bubbles refactor](https://github.com/motech/motech/pull/357/files) the old string manipulations directives were removed and new functionality wasn't added to filters.

### DIFF
--- a/modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
+++ b/modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
@@ -731,11 +731,11 @@
         $scope.getPopoverType = function(filter) {
             if (!filter.manipulations || !Array.isArray(filter.manipulations)) {
                 if (filter.displayName) {
-                    var manipulations, mans;
-                    mans = filter.key.split('?');
-                    mans.shift();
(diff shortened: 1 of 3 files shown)


modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -729,6 +729,24 @@
         };
 
         $scope.getPopoverType = function(filter) {
+            if (!filter.manipulations || !Array.isArray(filter.manipulations)) {
+                if (filter.displayName) {
+                    var manipulations, mans;
```
kmadej: please change variable name 'mans'

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -549,8 +557,18 @@
                     return false;
                 }
                 element.on('click', function (event) {
-                    var modalScope;
-                    if (!$(event.target).hasClass('field-remove')){
+                    var modalScope,
+                        fun;
```
kmadej: a lot of fun

PR #581, opened 2016-06-09
MOTECH-2601: Make Scheduler support the MotechEvent metadata map

1stmateusz:


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -786,7 +797,7 @@ private void validateJob(JobKey key) throws SchedulerException {
 
         JobDataMap map = detail.getJobDataMap();
 
-        if (map != null && !map.getBooleanValue(UI_DEFINED)) {
+        if (map != null && map.get(EVENT_METADATA) != null && !(Boolean) ((Map<String, Object>) map.get(EVENT_METADATA)).get(UI_DEFINED)) {
```
kmadej: please here use also getBooleanValue, it should simplify the code a bit

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -600,15 +603,19 @@ private void scheduleRepeatingJob(RepeatingSchedulableJob job, boolean update) {
             jobRepeatCount = MAX_REPEAT_COUNT;
         }
 
+        Map<String, Object> metadata = new HashMap<>();
```
kmadej: please move this lines just before putting this map into jobDataMap

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
1stmateusz: It is impossible to use getBooleanValue here, because (Map<String, Object>) map.get(EVENT_METADATA) map that we are using here is a HashMap, not a JobDataMap, so there isn't getBooleanValue implemented.

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -643,6 +646,16 @@ private void scheduleRepeatingJob(RepeatingSchedulableJob job, boolean update) {
         scheduleJob(jobDetail, trigger, update);
     }
 
+    private Map<String, Object> createMetadataForMisfireSchedualbleJob(MisfireSchedulableJob job) {
```
kmadej: misspelling

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
kmadej: ok understand, please move this condition into method, because it is hard to understand at the first look

modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerServiceImpl.java
```diff
@@ -786,12 +800,16 @@ private void validateJob(JobKey key) throws SchedulerException {
...
         JobDataMap map = detail.getJobDataMap();
 
-        if (map != null && !map.getBooleanValue(UI_DEFINED)) {
+        if (map != null && isJobUIDefined((Map<String, Object>) map.get(EVENT_METADATA))) {
             throw new MotechSchedulerException(String.format("Job is not ui defined:\n %s\n %s", key.getName(),
                     key.getGroup()));
         }
     }
 
+    private Boolean isJobUIDefined(Map<String, Object> jobMetadataMap) {
```
kmadej: please return primitive type here, additional this method should have jobdatamap parameter, all this conditions in if should be here to improve code readability

kmadej:
test this please

PR #582, opened 2016-06-13
MOTECH-2482: tasks module placeholder fix

1stmateusz:
disallow placeholder value to copy into form field


modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -409,6 +409,12 @@
             restrict: 'A',
             require: '?ngModel',
             link: function (scope, element, attrs, ngModel) {
+                var parent = scope;
+
+                while (parent.msg === undefined) {
```
pgesek: I would just do while (!parent.msg)

pgesek: Also, why are we checking for message, not for $parent? Seems very strange

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -428,12 +434,16 @@
...
                 element.bind('blur', function (event) {
+                    var content;
                     event.stopPropagation();
                     if(element[0] !== event.target){
                         return;
                     }
-                    ngModel.$setViewValue(readContent(element));
-                    scope.$apply();
+                    content = readContent(element);
+                    if (!(attrs.divPlaceholder && parent.msg(attrs.divPlaceholder).toLowerCase() === content.toLowerCase())) {
```
pgesek: why is the placeholder considered content? The proper fix would be to block this behaviour, the placeholder shouldn't be treated as input

modules/tasks/tasks/src/main/resources/webapp/js/directives.js
```diff
@@ -342,7 +342,7 @@
                         field = ele.data('value');
                         container.append(formatField(field));
                     }
-                }else{
+                } else if (!(this.tagName && this.tagName.toLowerCase() === 'em')) {
```
kmadej: please move this condition into separate method, it will be more readable. This method should be named for example 'isTagForPlaceholder'

PR #583, opened 2016-06-14
MOTECH-2622: Fixed concurrency issue in tasks

pgesek:
KeyEvaluator holds the current task state - the task context, yet it
was stored as a variable by the task action executor which is a
Sprint singleton. This would cause concurrency issues.


pgesek:
test this please

PR #593, opened 2016-06-20
MOTECH-2566 TypeError in javascript console browser - Cannot read pro…

wstrzelczyk:
MOTECH-2566 TypeError in javascript console browser - Cannot read property 'basic' of undefined

### DIFF
--- a/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
@@ -4433,7 +4433,7 @@
         $scope.isFieldSelected = function(name) {
             var i;
             for (i = 0; i < $scope.selectedFields.length; i += 1) {
-                if ($scope.selectedFields[i].basic.name === name) {
+                if ($scope.selectedFields[i] && $scope.selectedFields[i].basic.name === name) {
                     return true;
                 }
             }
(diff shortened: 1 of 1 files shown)


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4433,7 +4433,7 @@
         $scope.isFieldSelected = function(name) {
             var i;
             for (i = 0; i < $scope.selectedFields.length; i += 1) {
-                if ($scope.selectedFields[i].basic.name === name) {
+                if ($scope.selectedFields[i] !== undefined && $scope.selectedFields[i].basic.name === name) {
```
ngraczewski: Wouldn't "$scope.selectedFields[i] && $scope.selectedFields[i].basic.name === name" be enough?

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4444,7 +4444,7 @@
             var i, field;
             for (i = 0; i < $scope.allEntityFields.length; i += 1) {
                 field = $scope.allEntityFields[i];
-                if (field.basic.name === name) {
+                if (field.basic !== undefined && field.basic.name === name) {
```
ngraczewski: Wouldn't "$scope.basic && field.basic.name === name" be enough?

nickdotreid: @ngraczewski 
If field.basic is undefined and you try to call a property of it, javascript will throw an error and not evaluate the expression (or the rest of the function)

Yea, I know it looks stupid...

ngraczewski: Oh, I made a typo up there, I meant "field.basic" instead of "$scope.basic". My bad.

PR #596, opened 2016-06-23
MOTECH-1269 Sets proper data after reverting history

martokarski:


PiotrZurawskiSolDevelo:
I checked this, save button is working correctly with this changes.

PR #600, opened 2016-06-24
MOTECH-2601: fix retriving UI_DEFINED value from job

1stmateusz:


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/service/impl/MotechSchedulerDatabaseServiceImpl.java
```diff
@@ -494,8 +495,8 @@ private String getJobStatus(TriggerKey triggerKey) throws SchedulerException {
     private boolean isUiDefined(byte[] bytes) throws IOException, ClassNotFoundException {
         try (InputStream is = new ByteArrayInputStream(bytes);
              ObjectInputStream ois = new ObjectInputStream(is)) {
-            JobDataMap dataMap = (JobDataMap) ois.readObject();
-            return dataMap.getBoolean(UI_DEFINED);
+            JobDataMap jobDataMap = (JobDataMap) ois.readObject();
+            return jobDataMap.get(SchedulerConstants.EVENT_METADATA) != null &&  (Boolean) ((Map<String, Object>) jobDataMap.get(SchedulerConstants.EVENT_METADATA)).get(UI_DEFINED);
```
kmadej: please add this condition to private method it will be more readable.

Please add a proper condition to IT in separate PR

PR #611, opened 2016-07-04
MOTECH-2601: fix scheduling repeating period job

1stmateusz:
- also changed removing scheduler parameters from event metadata


modules/scheduler/scheduler/src/main/java/org/motechproject/scheduler/listener/MotechSchedulerListener.java
```diff
@@ -39,11 +38,14 @@ public void handleScheduleRepeatingJobEvent(MotechEvent event) {
         Map<String, Object> metadata = event.getMetadata();
 
         String jobSubject = (String) metadata.get(JOB_SUBJECT);
+        parameters.remove(JOB_SUBJECT);
         Integer repeatCount = (Integer) metadata.get(REPEAT_COUNT);
+        parameters.remove(REPEAT_COUNT);
         Integer repeatIntervalInSeconds = (Integer) metadata.get(REPEAT_INTERVAL_TIME);
+        parameters.remove(REPEAT_COUNT);
```
kmadej: something wrong here and why do you remove this from parameters ? I assume it should be only in metadata

PR #623, opened 2016-07-12
MOTECH-2074 Bulk delete in MDS UI

srujan92:
1. Created controllers at backend to delete bulk selected instances.
2. Added a checkbox option for entity instances table, for multiple selecting.
3. Created a UI model to delete selected instances.


motech-gerrit:

motech-gerrit:

motech-gerrit:

motech-gerrit:

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3512,6 +3533,11 @@
             $('#importInstanceModal').modal('show');
         };
 
+        $scope.deleteSelectedInstances = function() {
+            $scope.checkedEntitiesJSON = $scope.getCheckedEntitiesJSON();
+            $("#deleteSelectedInstancesModal").modal('show');
```
nickdotreid: Ideally this modal would use BootstrapDialog3 and would be tied into the modal window sequencing // but this will work for now, and we will change this when the MDS module is moved to MOTECH-UI

platform/mds/mds-web/src/main/resources/webapp/messages/messages.properties
```diff
@@ -224,8 +224,10 @@ mds.regex.dateTimeInfo=DateTime formats: yyyy[/.-]mm[/.-]dd hh:mm:ss
 mds.delete.deleteEntityTitle=Delete Entity
 mds.delete.deleteEntity=Selected entity and all its metadata will be deleted
 mds.delete.deleteInstanceTitle=Delete Instance
+mds.delete.deleteSelectedInstancesTitle=Delete Selected Instances
 mds.delete.deleteBlobContentTitle=Delete BLOB content
 mds.delete.deleteInstance=Selected instance will be deleted. Proceed?
+mds.delete.deleteSelectedInstances=All the selected instances will be deleted. Proceed?
```
nickdotreid: Can we add the number of selected instances into this string? Its confusing to read this message...

nickdotreid:
So, I noticed that I can't delete rows from the permissions table... but I could delete my platform settings (and therefore broke my MOTECH installation)

The hit area on the checkboxes is very small -- which is a UX problem I'm willing to deal with later.

The delete button needs to be disabled when no checkboxes are selected... or the confirm delete modal should read "please select some items"

In this UI, how does someone else all the records in a table

and sorry for the slow code review

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
srujan92: If we are changing this modal to BootsrapDialog3, we might even need to change all the other modals to that type because that looks better. As you said we will change this MDS module is moved to MOTECH-UI or  I could create sub-task and do it now.

srujan92:
Sorry, I didn't do permission table, I only concentrated on instances table since the issue said, I will do it too.
I will also see regarding deleting the disable button when unselected.
About deleting all, the check box on the table heading row will select all and therefore the delete action will be performed on all instances.

nickdotreid:
@srujan92 When deleting all instances, it looks like only the instances on a single page are deleted, and not the entire table (I'm judging only by your code)

Is there a way we can keep the option to delete all records in the table across multiple pages?

srujan92:
@nickdotreid I have implemented all changes you suggested, it works fine with permissions table too. Please do check and reply at your convenience.

platform/mds/mds-web/pom.xml
```diff
@@ -53,6 +53,11 @@
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
```
kmadej: please use com.google.gson library

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/controller/InstanceController.java
```diff
@@ -139,6 +145,30 @@ public void deleteInstance(@PathVariable Long entityId, @PathVariable Long insta
...
+    @ResponseStatus(HttpStatus.OK)
+    public void deleteAllInstances(@PathVariable Long entityId) {
+        instanceService.deleteAllInstances(entityId);
+    }
+
+    @RequestMapping(value = "/instances/{entityId}/deleteSelected/{instanceIds}", method = RequestMethod.DELETE)
+    @ResponseStatus(HttpStatus.OK)
+    public void deleteSelectedInstances(@PathVariable Long entityId, @PathVariable String instanceIds) throws ParseException {
+        List<Long> instanceIdsList = new ArrayList();
+        JSONParser parser = new JSONParser();
```
kmadej: please use com.google.gson library

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/controller/InstanceController.java
```diff
@@ -52,9 +55,12 @@
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.Iterator;
+import java.util.ArrayList;
 
 import static org.apache.commons.lang.CharEncoding.UTF_8;
 
+
```
kmadej: remove empty line

platform/mds/mds-web/src/main/java/org/motechproject/mds/web/controller/InstanceController.java
```diff
@@ -139,6 +145,30 @@ public void deleteInstance(@PathVariable Long entityId, @PathVariable Long insta
...
+        JSONObject instanceIdsJsonObject = (JSONObject) instanceIdsJson;
+        Iterator iterator = instanceIdsJsonObject.keySet().iterator();
+        while (iterator.hasNext()) {
+            String key = (String) iterator.next();
+            String value =  instanceIdsJsonObject.get(key).toString();
+            instanceIdsList.add(Long.parseLong(value));
+        }
+
+        instanceService.deleteSelectedInstances(entityId, instanceIdsList);
+
```
kmadej: remove empty line

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3411,6 +3420,18 @@
...
+            var selectedId = {};
+            $('#instancesTable').find('input:checked').each(function(index){
+                selectedId[index] =  $(this).parent().parent().attr('id');
+            });
+            return JSON.stringify(selectedId);
+        };
+
+
+
+
```
kmadej: remove empty lines

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3512,6 +3535,8 @@
 
         $scope.customModals = [];
 
+
```
kmadej: remove empty lines

platform/mds/mds-web/src/main/resources/webapp/messages/messages.properties
```diff
@@ -224,8 +225,12 @@ mds.regex.dateTimeInfo=DateTime formats: yyyy[/.-]mm[/.-]dd hh:mm:ss
 mds.delete.deleteEntityTitle=Delete Entity
 mds.delete.deleteEntity=Selected entity and all its metadata will be deleted
 mds.delete.deleteInstanceTitle=Delete Instance
+mds.delete.deleteAllInstancesTitle=Delete All Instances
+mds.delete.deleteSelectedInstancesTitle=Delete Selected Instances
 mds.delete.deleteBlobContentTitle=Delete BLOB content
 mds.delete.deleteInstance=Selected instance will be deleted. Proceed?
+mds.delete.deleteAllInstances= All the Instances will be deleted. Proceed?
```
kmadej: 'Instances' should be lowercase

srujan92: @kmadej I have updated the changes, please do check.

platform/mds/mds-web/pom.xml
```diff
@@ -53,6 +53,11 @@
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.6.1</version>
```
kmadej: in modules and motech repo we are using com.google.code.gson from artifact 'org.motechproject.com.google.code.gson'. The version is always defined in main pom 'motech.pom or modules.pom', in child we are only defining groupId and artifactId. You can check the following poms to look how to correctly add dependency into this module :
 openmrs/pom.xml, message-campaign/pom.xml, schedule-tracking/pom.xml

PR #625, opened 2016-07-14
MOTECH-2720: Added XML parser library to server

mkruszynski:
On server was no XML parsing library. That's why commcare request was causing response code 500.

### DIFF
--- a/platform/server/pom.xml
+++ b/platform/server/pom.xml
@@ -212,10 +212,10 @@
             <artifactId>motech-tomcat-it</artifactId>
             <version>${project.version}</version>
         </dependency>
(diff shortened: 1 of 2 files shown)


platform/server/pom.xml
```diff
@@ -212,6 +212,11 @@
             <artifactId>motech-tomcat-it</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>xml-apis</groupId>
+            <artifactId>xml-apis</artifactId>
+            <version>1.4.01</version>
```
pgesek: Version should be specified in the dependency management section of the main pom

PR #629, opened 2016-07-15
MOTECH-2716: Fixed modal blocking

mkruszynski:
Bootstrap div (covering whole page), which is always before our splash, have too low z-index. Bootstrap doesn't see previously called modal, and is setting too low value to covering div. Splash modal has right z-index.
These changes are changing only z-index value of blocking div.


sebbrudzinski:
test this please

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -87,6 +87,8 @@
                 dialog.getModalBody().css('padding', '0');
 
                 dialog.open();
+
+                dialog.getModal().prev().css('z-index', dialog.getModal().css('z-index'));
```
sebbrudzinski: This seems like a big hack, not a fix. What are we referring to with "prev()" here? I assume this will depend on the place it's called from? Why is the problem only appearing in some places?

sebbrudzinski: This seems like a big hack, not a fix. What are we referring to with "prev()" here? I assume this will depend on the place it's called from? Why is the problem only appearing in some places?

mkruszynski: @sebbrudzinski This is hack, indeed, but there's no other way to fix it. Problem appears when we call more than 1 modal on page. Every modal is preceded by div which is covering whole page - always. Bootstrap is setting too low z-index to this div. Without manual fix of z-index the new modal will be always on the same z-index with previous modal.

sebbrudzinski: @mkruszynski What kind of modal is used in Tasks? Doesn't it use our own ModalService, that was responsible for displaying only the latest modal?

mkruszynski: @sebbrudzinski e.g. in task import we call first modal using html arguments (like here: http://www.w3schools.com/bootstrap/bootstrap_ref_js_modal.asp) and then second using LoadingModal. Both are using BootstrapDialog, but LoadingModal is checking that only one instance is opened.

sebbrudzinski: So it doesn't use our own ModalService? If so, it probably should and it should also fix the problem.

nickdotreid: I agree to @sebbrudzinski, we should be using our ModalService

mkruszynski:
@sebbrudzinski @nickdotreid Both modals are created by our service now. Could you please take a look at this now.

sebbrudzinski:
test this please

platform/server-bundle/src/main/resources/webapp/js/services.js
```diff
@@ -14,12 +14,22 @@
         LoadingModal.closeEvent($rootScope, function(event) { that.show(); });
```
ngraczewski: As far as I understand after a talk with Marcin all the modifications here were made to prevent a modal from being opened multiple times. I think we should find the reason of such behavior rather than hack around it.

PR #643, opened 2016-08-02
MOTECH-2784 Tasks getTrigger in GET

filgomul:
Changed URL to use GET, not POST, to fetch data from the server

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/web/ChannelController.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/web/ChannelController.java
@@ -1,15 +1,16 @@
 package org.motechproject.tasks.web;
 
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import org.motechproject.tasks.domain.mds.channel.Channel;
-import org.motechproject.tasks.domain.mds.task.TaskTriggerInformation;
 import org.motechproject.tasks.domain.mds.channel.TriggerEvent;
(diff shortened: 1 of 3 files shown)


motech-gerrit:

motech-gerrit:

motech-gerrit:

martokarski:
test this please

martokarski:
I think, that parsing request param to JSON isn't the best idea. Could we change this?

martokarski:
test this please

modules/tasks/tasks/src/main/java/org/motechproject/tasks/web/ChannelController.java
```diff
@@ -98,9 +101,18 @@ public TriggersLists getTriggers(@RequestParam String moduleName,
      * @param info  the information about the trigger
      * @return the trigger matching given information
      */
-    @RequestMapping(value = "channel/trigger", method = RequestMethod.POST)
+    @RequestMapping(value = "channel/trigger", method = RequestMethod.GET)
     @ResponseBody
-    public TriggerEvent getTrigger(@RequestBody TaskTriggerInformation info) {
+    public TriggerEvent getTrigger(TaskTriggerInformation info) {
```
sebbrudzinski: Does this refactor even make sense? If we have got to provide a huge object as URL parameter, it would probably be better to leave it as POST payload.

filgomul:
As @sebbrudzinski said, if we have got to pass a huge object through URL, it may cause some problems, so we will close this ticket and do not merge the changes.

PR #648, opened 2016-08-04
MOTECH-2785: Fixed unwanted blocking of the ui in MDS

MichalKus:
Fixed unwanted blocking of ui
Fixed modal which didn`t close when we go back to previous site

### DIFF
--- a/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
@@ -633,7 +633,7 @@
     /**
     * The MdsSchemaEditorCtrl controller is used on the 'Schema Editor' view.
     */
-    controllers.controller('MdsSchemaEditorCtrl', function ($scope, $timeout, $http, Entities, MDSUsers, Permissions, MDSUtils, Locale, ModalFactory, LoadingModal, $rootScope) {
+    controllers.controller('MdsSchemaEditorCtrl', function ($scope, $timeout, $http, Entities, MDSUsers, Permissions, MDSUtils, Locale, ModalFactory, LoadingModal) {
 
         MDSUtils.setCustomOperatorFunctions($scope);
 
--- a/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
@@ -2047,11 +2047,7 @@
             var exists;
 
             if ($scope.advancedSettings !== null && $scope.lookup !== undefined && $scope.lookup.lookupName !== undefined) {
-                $timeout(function() {
-                    LoadingModal.open();
-                    $scope.validateLookupName($scope.lookup.lookupName);
-                    LoadingModal.close();
-                }, 1000);
+                $scope.validateLookupName($scope.lookup.lookupName);
             }
         });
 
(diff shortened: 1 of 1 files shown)


motech-gerrit:

motech-gerrit:

motech-gerrit:

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2047,9 +2047,19 @@
             var exists;
 
             if ($scope.advancedSettings !== null && $scope.lookup !== undefined && $scope.lookup.lookupName !== undefined) {
```
wstrzelczyk: @sebbrudzinski We really need to have LoadingModal in this place?

jredlarski:
test this please

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
jredlarski: @sebbrudzinski @MichalKus I believe that we can get rid of it here. If so, I think the timeout should be removed as well and I would left just the validation of the lookup name

jredlarski:
test this please

jredlarski:
@kmadej Looks good to me

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -2047,9 +2047,15 @@
...
-                LoadingModal.open();
                 $scope.validateLookupName($scope.lookup.lookupName);
-                LoadingModal.close();
+            }
+        });
+
+
+        $rootScope.$on('$stateChangeStart', function(event, toState, toParams, fromState, fromParams) {
```
wstrzelczyk: @jredlarski @MichalKus Do we really need catch every change of state in this place?

jredlarski: @wstrzelczyk @MichalKus We need to catch the change of state from `mds.schemaEditor` when `#advancedObjectSettingsModal` is open to force it to be closed on such state change. I can see that we catch the change of state only from `mds.schemaEditor` with either modal being open or closed

jredlarski:
@MichalKus Instead of using method which is listening for the change of state, let's block the ability to click outside the advanced settings modal.

jredlarski:
test this please

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -633,7 +633,7 @@
     /**
     * The MdsSchemaEditorCtrl controller is used on the 'Schema Editor' view.
     */
-    controllers.controller('MdsSchemaEditorCtrl', function ($scope, $timeout, $http, Entities, MDSUsers, Permissions, MDSUtils, Locale, ModalFactory, LoadingModal) {
+    controllers.controller('MdsSchemaEditorCtrl', function ($scope, $timeout, $http, Entities, MDSUsers, Permissions, MDSUtils, Locale, ModalFactory, LoadingModal, $rootScope) {
```
jredlarski: @MichalKus Please remove `$rootScope` from this controller if it is not being used anywhere inside it

jredlarski:
test this please

jredlarski:
@kmadej Looks good to me

PR #659, opened 2016-08-18
MOTECH-2831: Fixed Lookup return multiple entities

mkruszynski:


platform/mds/mds-web/src/main/resources/webapp/js/directives.js
```diff
@@ -2313,7 +2313,7 @@
                         values: {
                             path: advancedPath,
                             advanced: true,
-                            value: [value]
+                            value: [!value]
```
kmadej: this small change fixes the problem ? I checked in git history and this line has not been changed, however this feature has recently worked properly, so I suppose that the real problem is in different place. Can you look deeper into this bug ?

PR #661, opened 2016-08-18
MOTECH-2840: 500 status code after entering module settings

jhopen:


modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/SettingsServiceImpl.java
```diff
@@ -244,24 +244,25 @@ public void saveRawFile(MultipartFile file, String filename, long bundleId) {
...
-            InputStream is = null;
-            URL url = enumeration.nextElement();
-            try {
-                is = url.openStream();
-                Properties defaultBundleProperties = new Properties();
-                defaultBundleProperties.load(is);
-                if (!url.getFile().isEmpty()) {
-                    //We want to store plain filename, without unnecessary slash prefix
-                    allDefaultProperties.put(url.getFile().substring(1), defaultBundleProperties);
+        if(enumeration != null) {
```
mkruszynski: Use CollectionUtils.isNotEmpty

mkruszynski:
@kmadej it looks good to me

modules/admin/src/main/java/org/motechproject/admin/internal/service/impl/SettingsServiceImpl.java
```diff
@@ -244,24 +246,25 @@ public void saveRawFile(MultipartFile file, String filename, long bundleId) {
...
-            InputStream is = null;
-            URL url = enumeration.nextElement();
-            try {
-                is = url.openStream();
-                Properties defaultBundleProperties = new Properties();
-                defaultBundleProperties.load(is);
-                if (!url.getFile().isEmpty()) {
-                    //We want to store plain filename, without unnecessary slash prefix
-                    allDefaultProperties.put(url.getFile().substring(1), defaultBundleProperties);
+        if (CollectionUtils.isNotEmpty(Collections.list(enumeration))) {
```
sebbrudzinski: @jhopen or @mkruszynski  Can you explain why adding this if solves all our problems? It seems that if this enumeration is empty but not null, it should just skip the while loop below and return an empty map. Can this be null? If yes, when? and is this a correct behaviour?

jhopen: method findEntries returns an enumeration for each pattern matching entry or null if there is no match.
For modules IHE, mTraining and Event-Logging there are only raw data as mentioned in ticked so method should return null.

sebbrudzinski: Thanks @jhopen  Do we currently support editing of the raw module data via the Admin module UI? (Either old one or the new one)

jhopen: Yes, in both old and new UI raw data can be edited

PR #667, opened 2016-08-24
MOTECH-2860 Missing alerts with messages about validation errors

wstrzelczyk:
MOTECH-2860 Missing alerts with messages about validation errors in MDS module
- Restored error messages in MDS module

### DIFF
--- a/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
+++ b/platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
@@ -3563,13 +3563,12 @@
             });
         };
 
-        $scope.ngFormNameAttrSuffix = "form";
 
         $scope.addNewRelatedInstance = function (field) {
             $scope.relatedMode.isNested = true;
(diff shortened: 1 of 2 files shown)


platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -3563,13 +3563,12 @@
             });
         };
 
-        $scope.ngFormNameAttrSuffix = "form";
```
pgesek: From what I recall this suffix was added because of issues with the name duplicating based on the name of the field. Should this get removed?

wstrzelczyk:
The problem with duplicate names existed because generated form, requires the prior compilation of the field name (the directive doing it).

Adding this suffix to each field name does not change anything.

pgesek:
Does this change solve the problem?

wstrzelczyk:
Yes, this change should fixed that problem.

PR #669, opened 2016-08-25
MOTECH-2859 Can't read field bubble in Key Value map

piopawel:


motech-gerrit:

motech-gerrit:

motech-gerrit:

sebbrudzinski:
test this please

modules/tasks/tasks/src/main/resources/webapp/css/tasks.css
```diff
@@ -254,6 +254,7 @@ manipulation-sorter .manipulation .remove:before {
 
 .tasks .map-field {
     width: 30%;
+    overflow-x: auto;
```
sebbrudzinski: The Tasks module currently uses auto-expandable inputs (at least for raw text values). Could you verify whether we could somehow auto-expand the inputs for bubbles as well? This would look way better than adding a scroll.

sebbrudzinski:
test this please

sebbrudzinski:
Why do bubbles expand to fill the whole available space when I drop them on map key or value input?

piopawel:
@sebbrudzinski The CSS property should be max-value not value. I've already corrected it.

sebbrudzinski:
test this please

sebbrudzinski:
@piopawel  For future reference, the commit and pull request titles should contain a short description of work that was introduced, rather than ticket title.

PR #671, opened 2016-08-29
MOTECH-2757: Fixed admin queues messages

mkruszynski:


mkruszynski:
test this please

modules/admin/src/main/resources/webapp/js/app.js
```diff
@@ -108,7 +108,7 @@
                }
             })
             .state('admin.browse', {
-               url: '/browse',
+               url: '/browse?queueName',
```
sebbrudzinski: Questions:
1. Does it correctly display the data of the **selected** queue?
2. Should this be adjusted for topics as well?

mkruszynski: @sebbrudzinski 
1. Yes, in my MOTECH I had two queues without messages, and after click on queue name, list was empty, and third with something about 50 messages. Everything seems to work.

2. In topics doesn't exists possibility of displaying details.

PR #675, opened 2016-08-31
MOTECH-2728  MOTECH Home page

wstrzelczyk:


platform/server-bundle/src/main/resources/webapp/messages/messages.properties
```diff
@@ -181,4 +181,36 @@ logger.log.level.trace=TRACE
...
 logger.log.level.warn=WARN
-logger.log.level.error=ERROR
\ No newline at end of file
+logger.log.level.error=ERROR
+
+server.homepage.tab=Homepage
+server.homepage.title=Welcome to MOTECH
+
+server.homepage.dataServices.title=Manage Data
+server.homepage.dataServices.description=Use MOTECH Data Services to view, create, modify, or delete your application''s data and custom object definitions.
```
ngraczewski: application's

sebbrudzinski: this is correct

PR #678, opened 2016-09-02
MOTECH-2865: Travis integration tests on mysql and postgresql

mkruszynski:


.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+  - sudo service tomcat7 stop
+  - echo 'CATALINA_OPTS="-Xms512m -Xmx512m"' | sudo tee --append /usr/share/tomcat7/bin/setenv.sh
+  - sudo chown -R tomcat7:tomcat7 /var/lib/tomcat7/ /usr/share/tomcat7/ /var/log/tomcat7/
+
+before_script:
+  - sudo chmod +x ./travis_before_build.sh
+  - ./travis_before_build.sh
+
+script:
+  - mvn clean install -PIT -q
```
sebbrudzinski: Despite this, I can see that the Travis console outputs the following:
`mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V`
Can you verify where is that coming from?

mkruszynski: This is a default command in "install" phase of build. If we want to change this, we just should add e.g. "install: mvn clean install" to this file. 
Here it is in Travis documentation:
https://docs.travis-ci.com/user/languages/java/#Dependency-Management

sebbrudzinski: So do we separately run install and then tests now?

mkruszynski: Yes, I'm checking now a configuration with tests only. If they will succeed, I'll push a new configuration file.

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -y activemq -qq
+  - sudo ln -s /etc/activemq/instances-available/main /etc/activemq/instances-enabled/main
+  - sudo sed -e 's/<broker /<broker schedulerSupport="true" /' -i /etc/activemq/instances-enabled/main/activemq.xml
+  - sudo service activemq restart
+  - sudo apt-get install -y tomcat7
+  - sudo service tomcat7 stop
+  - echo 'CATALINA_OPTS="-Xms512m -Xmx512m"' | sudo tee --append /usr/share/tomcat7/bin/setenv.sh
+  - sudo chown -R tomcat7:tomcat7 /var/lib/tomcat7/ /usr/share/tomcat7/ /var/log/tomcat7/
```
sebbrudzinski: Can we move all of this to a script as well? Also, let's start naming them adequately to what they do (eg. set_up_databases.sh, install_dependencies.sh etc.)

mkruszynski: Of course, I'll handle this.

sebbrudzinski:
@mkruszynski  Can we close it now?

mkruszynski:
@sebbrudzinski Yes, I think this PR isn't more necessary. We have similar configuration in PR for MOTECH-2895.

PR #679, opened 2016-09-02
MOTECH-2867 Saving null if empty string in field

tomaszzawadzki:


motech-gerrit:

motech-gerrit:

motech-gerrit:

motech-gerrit:

platform/mds/mds-web/src/main/resources/webapp/js/controllers.js
```diff
@@ -4196,6 +4196,9 @@
 
             var values = $scope.currentRecord.fields;
             angular.forEach (values, function(value, key) {
+                if(value.type.defaultName === "str" && value.value === ""){
```
sebbrudzinski: Why are we extracting the default name parameter from the type for this check? Isn't there any more appropriate field?

tomaszzawadzki: defaultname
Data looks like this. If we want to check the type of field if it is string type i think we need this defaultName value.

sebbrudzinski: What about typeClass? It clearly indicates a string field.

tomaszzawadzki: Yes, typeClass looks better. I will change that.

sebbrudzinski:
test this please

PR #681, opened 2016-09-07
MOTECH-2895: Migrate PR jobs to Travis

mkruszynski:


sebbrudzinski:
@mkruszynski  Let's only remove log4j.xmls from tests. This should be sufficient for our 4MB limit.

PR #695, opened 2016-09-12
MOTECH-2898 Add manipulation help in Tasks manipulation modals

wstrzelczyk:
MOTECH-2898 Add manipulation help in Tasks manipulation modals 
- changed html static help modal to service help modal


sebbrudzinski:
@wstrzelczyk Closing the modification help modal should bring me back to the manipulation window if I opened it from there.

PR #698, opened 2016-09-13
MOTECH-2861: Switched off datanucleus property detachAllOnCommit

afijal:
I've set datanucleus property detachAllOnCommit to false, to fix poor performance while using entities with SCO files, which is visible mainly in tasks module. Because of that, now we have to detach SCO files in mds data browser, while other fields' values should be fetched by setting the property retainValues to true. From now, MotechDataService should be used only in transaction. An instance shouldn't be updated outside transaction.
I also added two methods which can help dealing with persistent object outside transaction, however it is not recommended :

detachedCopy, this method detached instance,
detachedCopyAll, this method detached all given instances

Additionally I added possibility to run ITs with switched on detachAllOnCommit. You have to set detach env to 'detach-true', for example :
mvn clean install -PIT -Ddetach=detach-true

My PR is based on older [one](https://github.com/motech/motech/pull/470), unmerged, by @kmadej


kmadej:
test this please

PR #699, opened 2016-09-13
MOTECH-2861: Tasks refactor

afijal:
We are trying to switch off datanucleus property - detachAllOnCommit because of performance breakdown. There is one main problem with SCO (second-class object) fields. The values for the fields are null outside transaction after restarting motech. In our modules like mds or web-security we are using dto objects. This type of objects are parsed to JSON file. However in Tasks module we have returned domain objects and unfortunately after switching off the mentioned property we will loose the SCO field values. In this PR I refactored tasks module to use dto objects in tasks controllers.

The changes regard this PR :

-  created dto classes
-  created toDto methods in domain classes
-  created constructors with dto object parameter
-  created TaskWebService (this service contains methods which are used in controllers and return domain objects, the all methods are annotated with transactional annotation and after getting object, the object is converted to dto)
-  added transactional annotation to tasks services (removed doInTransaction method because of that change and change the way of autowiring connected services)
-  changed serialization/deserialization of tasks. In order to keep tasks compatibility with previous motech version, the task config deserializer has two options. One to create objects from old jsons (with DataSource and FilterSet types of TaskConfigStep) and second from new jsons(with DataSourceDto and FilterSetDto types of TaskConfigStepDto)
-  moved enums from domain to domain.enums package because of package cycle detection between domain and dto
-  adjusted unit and integration tests to the new changes

PR description is taken from @kmadej [older PR](https://github.com/motech/motech/pull/466), as mine is just an adaptation of his changes to newer version of Motech


kmadej:
test this please

PR #707, opened 2016-09-16
MOTECH-2936: Fix Text Limits on OpenMRS Observations

palbecki:
There is a problem with postgres after change column "VALUE" type to TEXT.

> Failed to generate new Mapping of type org.datanucleus.store.rdbms.mapping.java.StringMapping, exception : SQL type **text** declared for field "org.motechproject.tasks.domain.mds.task.TaskActionInformation.values" of java type java.lang.String cant be mapped for this datastore.

Instead of text I have used VARCHAR with max length 20000. According to [mysql documentation](http://dev.mysql.com/doc/refman/5.5/en/string-type-overview.html):

> The effective maximum length of a VARCHAR is subject to the maximum row size (65,535 bytes, which is shared among all columns) and the character set used. For example, utf8 characters can require up to three bytes per character, so a VARCHAR column that uses the utf8 character set can be declared to be a maximum of 21,844 characters.

In postgreSQL max length for type varchar is 10485760.


motech-gerrit:

motech-gerrit:

modules/tasks/tasks/src/main/resources/db/migration/default/V2__MOTECH-2936.sql
```diff
@@ -0,0 +1,3 @@
+-- Changing column 'VALUE' from VARCHAR(500) to TEXT
```
kmadej: change comment

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/TaskActionInformation.java
```diff
@@ -28,6 +28,7 @@
 public class TaskActionInformation extends TaskEventInformation {
 
     private static final long serialVersionUID = -132464255615128442L;
+    private static final int MAX_VALUE_LENGTH = 20000;
```
kmadej: In the ticket description there is info about 'unlimited', so I suppose we should set this as highest as possible. Mysql varchar value is lower, so please set to this value

PR #715, opened 2016-09-27
Fixed retry task errors.

pkornowski:
During as conversion task activity to dto we didn't include activity ID.


kmadej:
It seems that you solve issue not connected with this ticket. Probably after recently Tasks migration this bug occurs, however it is not related with problem described in the ticket. Please change the topic of PR and try to reproduce the problem from JIRA

PR #718, opened 2016-09-30
MOTECH-2927: Hover on Bubble Should Show Full Path

0nat:
Fixed the way of displaying text that it fit in the bubble.


mkruszynski:
@kmadej It looks good to me.

PR #727, opened 2016-10-06
Documented REST end-points of Motech modules to support Swagger docs

shanmukhm:
Five modules are documented in this PR i.e. Scheduler, Admin, Tasks, Email and Server-bundle
* Added swagger annotations to document the end-points
* Using swagger-maven-plugin, swagger.json is generated for every module at build time
* Path to the json file is specified as restDocsPath of the particular module


kmadej:
Hi @shanmukhm 

please fix Task ITs and resolve conflicts

PR #729, opened 2016-10-10
MOTECH-2965: Added search on available Triggers List

MichalKus:
Searching triggers by name is now available.

### DIFF
--- a/modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
+++ b/modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
@@ -1346,33 +1346,37 @@
         };
 
         $scope.$watch('searchTrigger.displayName', function() {
-            var searchedTriggers = [];
+            var searchActive = true;
 
             if ($scope.hasStaticTriggers) {
-                $scope.staticTriggers.page = 1;
-                if ($scope.searchTrigger.displayName !== "") {
-                    searchedTriggers = $filter('filter')($scope.staticTriggers.triggers, $scope.searchTrigger);
-                    $scope.staticTriggers.total = parseInt(searchedTriggers.length / $scope.pageSize, 10);
-                    if (searchedTriggers.length % $scope.pageSize > 0) {
-                        $scope.staticTriggers.total += 1;
-                    }
-                } else {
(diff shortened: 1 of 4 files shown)


pkornowski:
Please, use here dynamicTriggersCount instead of PAGE_SIZE
https://github.com/motech/motech/pull/729/commits/c65e97bfc521ed8400adfcb4e677e13fa3a2ffc0#diff-d1c18e008c57d03e2312c8fdfabb6f32R82

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1332,27 +1332,48 @@
...
+                    searchedTriggers = $filter('filter')($scope.dynamicTriggers.triggers, $scope.searchTrigger);
+                    $scope.dynamicTriggers.total = parseInt(searchedTriggers.length / $scope.pageSize, 10);
+                    if (searchedTriggers.length % $scope.pageSize > 0) {
+                        $scope.dynamicTriggers.total += 1;
+                    }
+                } else {
+                    $scope.dynamicTriggers.total = $scope.dynamicTriggersPages;
+                }
+            }
+        });
```
kmadej: this two if conditions look almost the same, the only change is dynamic/static. I think it can be done simplier

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1346,34 +1346,38 @@
...
-                    searchedTriggers = $filter('filter')($scope.dynamicTriggers.triggers, $scope.searchTrigger);
-                    $scope.dynamicTriggers.total = parseInt(searchedTriggers.length / $scope.pageSize, 10);
-                    if (searchedTriggers.length % $scope.pageSize > 0) {
-                        $scope.dynamicTriggers.total += 1;
-                    }
-                } else {
-                    $scope.dynamicTriggers.total = $scope.dynamicTriggersPages;
+                searchActive = $scope.changePagesAfterSearch($scope.dynamicTriggers, searchActive);
+                if(searchActive === false) {
+                    $scope.dynamicTriggers.total = $scope.staticTriggersPages;
```
kmadej: should be dynamicTriggersPages

PR #731, opened 2016-10-12
MOTECH-2901: Added ability to lookup 'null' values in a task data source.

pkornowski:


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/util/KeyEvaluator.java
```diff
@@ -64,13 +64,17 @@ public String evaluateTemplateString(String template) throws TaskHandlerExceptio
 
         for (KeyInformation key : parseAll(template)) {
             Object value = getValue(key);
-            String stringValue = value != null ? value.toString() : "";
```
kmadej: just change "" to null

PR #733, opened 2016-10-13
MOTECH-2989: Added better indication that Task was Run and Filtered

MichalKus:
When task was triggered but not executed because of filter, there is proper information in task history displayed.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskActivityService.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskActivityService.java
@@ -24,7 +24,7 @@ public interface TaskActivityService {
     long addTaskStarted(Task task, Map<String, Object> parameters);
 
     /**
-     * If task was filtered and not executed, it marks the activity as "FILTERED".
+     * Marks the activity as "FILTERED", if task was filtered and not executed.
      *
      * @param activityId the id of the activity
      */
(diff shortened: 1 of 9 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/TaskActivityService.java
```diff
@@ -24,6 +24,13 @@
     long addTaskStarted(Task task, Map<String, Object> parameters);
 
     /**
+     * If task was filtered and not executed, it marks the activity as "FILTERED".
```
kmadej: the first word in method javadoc should be verb in 3rd person

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -48,6 +49,16 @@ public long addTaskStarted(Task task, Map<String, Object> parameters) {
 
     @Override
     @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void addTaskFiltered(Long activityId) {
```
kmadej: add unit test to TaskActivityServiceImplTest

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskTriggerHandler.java
```diff
@@ -176,6 +176,8 @@ private void handleTask(Task task, Map<String, Object> parameters, boolean isRet
                 for (int i = 0; i < task.getActions().size(); i++) {
                     executor.execute(task, task.getActions().get(i), i, taskContext, activityId);
                 }
+            } else {
```
kmadej: please add test to TaskTriggerHandlerTest

PR #735, opened 2016-10-17
MOTECH-2859 Enhanced adjustText function and CSS styles in specific situations

wstrzelczyk:
Enhanced adjustText function and CSS styles in specific situations


modules/tasks/tasks/src/main/resources/webapp/partials/field.html
```diff
@@ -2,7 +2,7 @@
 <span ng-if="!editable && field.providerName" class="field-bubble field-{{field.prefix}}" data-trigger="hover" data-placement="top" bs-popover="{{field.prefix}}.{{field.providerName}}.{{field.providerType}}#{{field.objectId}}.{{field.fieldKey}}">{{displayName}}</span>
 <span ng-if="!editable && !field.providerName && !field.eventKey" class="field-bubble field-{{field.prefix}}">{{displayName}}</span>
 <span ng-if="editable" class="field-bubble field-{{field.prefix}}" manipulation-modal manipulation-type="field.type" manipulations="field.manipulations">
+    <button class="field-remove close badge-close"></button>
```
kmadej: why do you remove text from button ?  Additionally it should be in message.properties instead of raw text

wstrzelczyk: I removed this label text because it never display, we display only 'x' sign

kmadej: ok

PR #738, opened 2016-10-20
MOTECH-2970 Added the possibilty of naming a Data Source or Task Action.

pkornowski:


modules/tasks/tasks/src/main/resources/webapp/partials/form-data-source.html
```diff
@@ -18,6 +18,12 @@
     </div>
     <div id="collapse-step-{{step.order}}" class="panel-collapse stripes-bg collapse in overflowChange" style="overflow: visible;">
         <div class="panel-body inside">
+            <div class="form-group control-section" >
+                <label class="col-md-2 col-sm-3 control-label task-label">{{msg('task.ds.name')}}</label>
+                <div class="col-md-8 col-sm-6">
```
kmadej: task.datasource.displayName

modules/tasks/tasks/src/main/resources/webapp/partials/form-data-source.html
```diff
@@ -18,6 +18,12 @@
...
+                    <input class="form-control" type='text' ng-change="addSpecifiedDSName(step)" ng-model='step.specifiedName' id="step.specifiedName">
+                </div>
```
kmadej: instead of DS use DataSource

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/DataSource.java
```diff
@@ -32,6 +32,7 @@
     private Long objectId;
     private String type;
     private String name;
+    private String specifiedName;
```
kmadej: can we name this displayName ?

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/TaskActionInformation.java
```diff
@@ -106,13 +109,14 @@ public TaskActionInformation(String displayName, String channelName, String modu
      * @param serviceMethod    the task service method
      * @param values           the map of values
      */
-    public TaskActionInformation(String name, String displayName, // NO CHECKSTYLE More than 7 parameters (found 9).
+    public TaskActionInformation(String name, String specifiedName, String displayName, // NO CHECKSTYLE More than 7 parameters (found 9).
```
kmadej: constructor where we are assigning values should be the last one

modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/DataSource.java
pkornowski: It is possible however it will take a lot of time, because on the frontend, objects action and datasource have already different field named 'displayName'

kmadej:
The build passed on travis, there is some problem with communication between travis and github. I am merging this PR without travis status

PR #741, opened 2016-10-25
Fixed manipulations and drag/drop bubble on the key-value map.

pkornowski:


mkruszynski:
@pkornowski Looks good to me.

PR #743, opened 2016-10-26
MOTECH-2970: Fixes after QA.

pkornowski:
- Corrected displaying custom data source names in the filter dropdown.
- Corrected saving task actions name.


modules/tasks/tasks/src/main/java/org/motechproject/tasks/domain/mds/task/TaskActionInformation.java
```diff
@@ -204,6 +204,7 @@ public boolean equals(Object obj) {
 
         return Objects.equals(this.serviceInterface, other.serviceInterface) &&
                 Objects.equals(this.serviceMethod, other.serviceMethod) &&
+                Objects.equals(this.specifiedName, other.specifiedName) &&
```
kmadej: please add this field in hashCode and Equals methods to DataSource, TaskACtionInformationDto, DataSourceDto

PR #746, opened 2016-11-02
MOTECH-3006 Added provision to retry multiple tasks at once

shanmukhm:
* Created a separate view for tasks history
* Added filters based on progress, task name and date of execution
* Added a button to retry all the failed tasks in current page

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/repository/TaskActivitiesDataService.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/repository/TaskActivitiesDataService.java
@@ -1,5 +1,7 @@
 package org.motechproject.tasks.repository;
 
+import org.joda.time.DateTime;
+import org.motechproject.commons.api.Range;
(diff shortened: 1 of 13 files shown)


kmadej:
@mjazdzdewski please pull this changes and check if feature works correctly

mjazdzdewski:
I've found few issues:
1. Sometimes (usually after a recent task execution) the history page fails to load completly. It loads all the records, but fails to convert it to user friendly informations.
1_load_fail
2. When I filter tasks so it shows only some of the failed tasks when I use Retry all failed Tasks button MOTECH starts to fail: cannot load any data or properly switch page. It basically breaks [down.]
3. When I switch to History page it lacks grey font text above the tabs which shows in other pages of the Task module.
3_missing_grey_text
4. I have also some doubts about how Retry all failed Tasks works. More details can be found in JIRA comment by @kmadej .

kmadej:
thanks @shanmukhm for corrections, @mjazdzdewski is now verifying this

mjazdzdewski:
It looks good to me.

PR #748, opened 2016-11-02
MOTECH-3023: Fixed 500 Error during first run

mkruszynski:


kmadej:
I suppose the problem is with datanucleus. I think that the best solution will be detached user object in findByUserName method

mkruszynski:
@kmadej I checked a solution with detached user object. The error was still occuring. This error exists only when user's roles are null. I checked and for user without any role, the read roles are an empty list, so null is some kind of incorrect read.

platform/web-security/src/main/java/org/motechproject/security/repository/MotechUsersDao.java
```diff
@@ -30,7 +30,17 @@
...
+            return null;
+        }
+
+        MotechUser retrievedUser = dataService.findByUserName(userName.toLowerCase());
+
+        if (retrievedUser != null && retrievedUser.getRoles() == null) {
+            return findByUserName(userName);
+        }
+
+        return retrievedUser;
```
kmadej: it is not the best solution. Try use getDetachedField to retrieve roles object

platform/web-security/src/main/java/org/motechproject/security/repository/MotechUsersDao.java
```diff
@@ -30,7 +30,17 @@
      */
     @Transactional
     public MotechUser findByUserName(String userName) {
-        return null == userName ? null : dataService.findByUserName(userName.toLowerCase());
+        if (userName == null) {
+            return null;
```
kmadej: please use only one return statement

platform/web-security/src/main/java/org/motechproject/security/repository/MotechUsersDao.java
```diff
@@ -30,7 +30,17 @@
...
+        }
+
+        MotechUser retrievedUser = dataService.findByUserName(userName.toLowerCase());
+
+        if (retrievedUser != null && retrievedUser.getRoles() == null) {
```
kmadej: please add comment to this code and provide ticket number on JIRA

PR #749, opened 2016-11-03
MOTECH-2998 Added disabling on save and corrected time window check

mjazdzdewski:


platform/commons-date/src/main/java/org/motechproject/commons/date/model/Time.java
```diff
@@ -210,6 +210,16 @@ public int compareTo(Time otherTime) {
     }
 
     /**
+     * Checks whether this is the same as the given time.
+     *
+     * @param other  the {@code Time} to be compared with this object
+     * @return true if this time is same as other, false otherwise
+     */
+    public boolean equals(Time other) {
```
kmadej: equals already exists

platform/commons-date/src/main/java/org/motechproject/commons/date/model/Time.java
```diff
@@ -234,10 +244,10 @@ public boolean isAfter(Time other) {
      *
      * @param start  the {@code Time} to be compared with this object
      * @param end  the {@code Time} to be compared with this object
-     * @return true if this time is between start and end, false otherwise
+     * @return true if this time is between start and end (inclusive), false otherwise
      */
     public boolean isBetween(Time start, Time end) {
-        return isAfter(start) && isBefore(end);
+        return isAfter(start) && isBefore(end) || equals(start) || equals (end);
```
kmadej: you should add isAfterOrEqual and so on

PR #754, opened 2016-11-09
MOTECH-3033: Added a possibility to copy/paste bubbles.

pkornowski:


modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1185,6 +1186,32 @@
             return newArray;
         };
 
+        $(document).click(function (event) {
```
wstrzelczyk: you should not capturing all click events on the whole document

modules/tasks/tasks/src/main/resources/webapp/js/controllers.js
```diff
@@ -1185,6 +1186,32 @@
...
+            if ($(event.target).hasClass("field-bubble")) {
+                $scope.lastSelectedField = $(event.target).parent().data("value");
+                $scope.$apply();
+            }
+        });
+
+        $(document).ready(function() {
+            var ctrlDown = false, ctrlKey = 17, cmdKey = 91, cKey = 67;
+
+            $(document).keydown(function(event) {
```
wstrzelczyk: you should not capturing all keydown, keyup events on the whole document

PR #775, opened 2016-11-25
MOTECH-3054: Added task filter check if string is null or blank

kmadej:


pkornowski:
@kmadej  Looks good to me.

PR #776, opened 2016-11-29
MOTECH-3052: Fixed: Action Filter Sets Should be Tied to a Specific A…

pkornowski:
…ction.

### DIFF
--- a/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
+++ b/modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
@@ -78,14 +78,12 @@ public class TaskActivityServiceImpl implements TaskActivityService {
     @Transactional(propagation = Propagation.REQUIRES_NEW)
     public void addFilteredExecution(Long activityId) {
         TaskActivity activity = taskActivitiesDataService.findById(activityId);
-        if (activity == null) {
-            return;
+        if (activity != null) {
+            TaskExecutionProgress progress = activity.getTaskExecutionProgress();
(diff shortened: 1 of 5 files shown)


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -69,18 +70,22 @@ public boolean addSuccessfulExecution(Long activityId) {
...
-            activity.getFields().clear();
+        return updateTaskProgress(progress, activity);
+    }
+
+    @Override
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void addFilteredExecution(Long activityId) {
+        TaskActivity activity = taskActivitiesDataService.findById(activityId);
+        if (activity == null) {
+            return;
```
kmadej: omit using return in void methods, it is less readable and it is hard to debugging

modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskActivityServiceImpl.java
```diff
@@ -200,6 +205,25 @@ public long getAllTaskActivitiesCount(Set<TaskActivityType> activityTypes, Range
         return taskActivitiesDataService.countByActivityTypesAndDate(activityTypes, dateRange);
     }
 
+    private boolean updateTaskProgress(TaskExecutionProgress progress, TaskActivity activity) {
+        boolean taskFinished = progress.getActionsSucceeded() + progress.getActionsFiltered() == progress.getTotalActions();
```
kmadej: '+' ? use && or ||

PR #785, opened 2016-12-09
MOTECH-3073: Fixed not updating table definitons.

kmadej:
Bug regarded changing required value and dropping columns


pkornowski:
@kmadej The code looks good to me.

PR #789, opened 2016-12-15
MOTECH-3088: Fixed filter edition in edit task

mkruszynski:


kmadej:
the same problem exists for string manipulation, please fix it

kkedys:
I checked and I find the errors: 

    Create a task, add a filter having the manipulations and save.
    Go to the task, add (or not) a manipulation and save.
    Go to the task (instead added manipulation is displayed: [task.undefined].

    Create a task, add a filter who does not the manipulations and save.
    Go to the task, add a manipulation and saved.
    Go to the task (previously added manipulations are not added).

kkedys:
@kmadej. I checked, the issue was fixed.

PR #791, opened 2016-12-19
MOTECH-3065: SQL migration schema update

mkruszynski:


kmadej:
Please change db migrations, because now this field types are mediumtext and bytea. Normally we should not change db migrations, however you added this recently, so you can do it

kmadej:
During creating instance for TaskError the following exception occurs :
Caused by: javax.jdo.JDODataStoreException: Insert of object "org.motechproject.tasks.domain.mds.task.history.TaskError__History@1c714deb" using statement "INSERT INTO MOTECH_TASKS_TASKERROR__HISTORY (creationDate,taskError__HistoryCurrentVersion,modificationDate,creator,ds,taskError__HistorySchemaVersion,modifiedBy,owner,message) VALUES (?,?,?,?,?,?,?,?,?)" failed : Data truncation: Data too long for column 'message' at row 1

kkedys:
@mkruszynski I checked and I found the errors concerning the manipulations:

1. Create a task, add a filter having the manipulations and save.
2. Go to the task, add (or not) a manipulation and save.
3. Go to the task (instead added manipulation is displayed: [task.undefined].

1. Create a task, add a filter who does not the manipulations and save.
2. Go to the task, add a manipulation and saved.
3. Go to the task (previously added manipulations are not added).

kkedys:
@mkruszynski When the user want delete the instances in TaskError and in StatusMessage in console 500 error appears. Please correct their Trash.

kkedys:
@kmadej, I checked,  it works properly.

PR #792, opened 2016-12-21
MOTECH-3090: Removed 'x' button in the modal where exist 'close' butt…

pkornowski:
…on in the task module.


kkedys:
@kmadej, I check it, the issue has been fixed.

PR #793, opened 2016-12-22
 MOTECH-3066: Removing whitespaces from action parameters

mkruszynski:


kkedys:
@kmadej, I checked this in different ways and I have not found any whitespaces in the JSON file.

PR #794, opened 2016-12-23
MOTECH-3081 Fixed the bug related to period modal in tasks

shanmukhm:


kkedys:
Checked. The Period modal works better, but when the user saved the action having a Period and enter again in this action, the Period is visible just in the field, when the user want see the modal window, the previous period disappears.

mkruszynski:
@kmadej I fixed occuring bugs in #798

PR #795, opened 2016-12-29
MOTECH-3074: Fixed: Newly Added Data Sources Don't Appear in Auto-Complete.

pkornowski:


kkedys:
@kmadej, I checked and now when the new data source is added or removed, the bubbles in actions are updated.

PR #796, opened 2016-12-30
MOTECH-3083 Handled passing null values to Lookup method

shanmukhm:


platform/mds/mds/src/main/java/org/motechproject/mds/query/PropertyBuilder.java
```diff
@@ -55,6 +55,9 @@ public static Property create(String name, Object value, String type, String ope
         } else if (value instanceof Range) {
             Range range = (Range) value;
             return new RangeProperty<>(name, range, type);
+        } else if (type.equals("java.lang.String") && StringUtils.isNotBlank(operator) && operator
```
kmadej: add this condition below into switch case statement and please use String.class.getName() method on the left side (raw string should always be on the left side for example : "test".equals(testVar)

platform/mds/mds/src/main/java/org/motechproject/mds/query/QueryUtil.java
```diff
@@ -141,6 +141,10 @@ public static String asMatchesPattern(String string) {
         return StringUtils.isNotEmpty(string) ? String.format(".*%s.*", string) : string;
     }
 
+    public static String asEqualsPattern(String string) {
+        return StringUtils.isNotEmpty(string) ? String.format("%s", string) : "";
```
kmadej: return StringUtils.defaultString(string);

kkedys:
@kmadej I check this issue and everything works correctly.

PR #797, opened 2017-01-09
MOTECH-3105: Fixed: Task Will Not Enable.

pkornowski:


modules/tasks/tasks/src/main/java/org/motechproject/tasks/service/impl/TaskServiceImpl.java
```diff
@@ -142,6 +142,18 @@
...
     @Override
     @Transactional
+    public Set<TaskError> setEnabledOrDisabled(final Task task) {
+        LOGGER.info("Saving task: {} with ID: {}", task.getName(), task.getId());
+        Set<String> fieldsToUpdate = new HashSet<>();
+        fieldsToUpdate.add("enabled");
+
+        tasksDataService.updateFromTransient(task, fieldsToUpdate);
+
+        return taskValidator.validate(tasksDataService.findById(task.getId()));
```
kmadej: we should drop validation, we are updating only one field and retrieving instance and validate it can take too much time

kkedys:
Checked. The ticket was difficult to verify from technical point of view, because we don't have the problem linked with the loading time during the loading of "enable" and "pause". I verified while it was loading on "network" in the console. The time of loading of the same task before the change was around 500ms and after the changes 50-70 ms. Aditionally a loading module has been added so it is impossible to cut loading time.

PR #798, opened 2017-01-12
MOTECH-3081: Fixed Period modal in Tasks UI

mkruszynski:


kkedys:
@kmadej I checked. The period modal works correctly.

PR #799, opened 2017-01-17
MOTECH-3110: Fixed DynamicChannelLoader logs

mkruszynski:


kkedys:
@kmadej I checked the error that appears in the logs after having importing the task that does not work (20161122_PrimeroEncontro_v14). Now instead of "Dynamic channel provider not found for module %s", "Dynamic channel provider not found for module [ the name of modules ]" appears.

PR #801, opened 2017-01-24
MOTECH-3112 Display Raw Syntax Does Not Work

srujan92:
Fixed the bug regarding display raw syntax in the tasks module


motech-gerrit:

kkedys:
@kmadej The option "Display raw syntax" works correctly. I tested on most modules.

PR #802, opened 2017-01-24
MOTECH-3111: Fixed: MOTECH Server URL is Reset on Server Restart

pkornowski:


platform/server-config/src/main/java/org/motechproject/config/service/impl/ConfigurationServiceImpl.java
```diff
@@ -127,7 +125,6 @@ public void save(BootstrapConfig bootstrapConfig) {
     }
 
     @Override
-    @Caching(cacheable = {@Cacheable(value = SETTINGS_CACHE_NAME, key = "#root.methodName") })
     public MotechSettings getPlatformSettings() {
```
kmadej: please try to add transactional, because this method is using object from db. Probably we tried it and for some reason it didnt work out, however please just try

kkedys:
@kmadej, the issue is fixed.

PR #805, opened 2017-01-27
MOTECH-3121: Hidden warnings in tasks history

mkruszynski:


kkedys:
@kmadej, the issue is fixed.

PR #806, opened 2017-01-30
MOTECH-2984: Added redirect to login page if session was destroyed.

pkornowski:


kkedys:
It depends where I click after the session finishes. Sometimes one see the login page immediately, sometimes one see the modal that is loading slowly and after that one see the login page. There are also other cases. One has to synchronize so that every time one can see the login screen immediately.

kkedys:
@mkruszynski, I verified the pull request, it doesn't matter where I click after the session time finishes, the user is redirected to login pages. There is one exception, when going in MDS module, the website blocks on MDS. Even when chosing another module, most likely the login page reappears. The problem is hard to reproduce, it appears from time to time. After consulting Kamil Madej, we agreed to merge it.

PR #807, opened 2017-02-02
MOTECH-3126: Fixed task-created entity instances

mkruszynski:


kmadej:
@mkruszynski it is bad solution. Really dangerous, because create method is used in many place. Please modify ActionHandlerService

kmadej:
It looks good to me, however Sheel added a comment to this ticket, probably we should remove all transactions from this class

pkornowski:
@kmadej 
Looks good to me. Now it works correctly.

PR #812, opened 2017-02-10
MOTECH-3130: Added warnings when data source unavailable

mkruszynski:


kkedys:
@mkruszynski  The changes are visible, one has to change some things: 
1. When the ID is downloaded in the trigger:
- it would be better if the warning of the action appears in red (so it will be unified)
- after having invade on the warnings, the following information appears: "Data input contains fields that are not defined in the trigger". This information should be changed because it may relate to data source or trigger (it has a link with point 2)

2. When the ID is downloaded of the data source: 
- information: "Object not found. Lookups are not available." has to appear in red or in the red field.
- in the action, one has to add a warning like it is in the case of when the ID is downloaded from trigger.

kkedys:
@mkruszynski please add the warning in action when the ID is download from data source.

kmadej:
There are still some issues :
1. If task contains trigger and data source with wrong case name, the warning does not appear on trigger
2. The trigger bubbles used in DS do not have triangle warning
3. After deleting outdated DS, the bubbles used in actions lost triangle warning and you can even save the task

kkedys:
@kmadej Everything works correctly.

PR #814, opened 2017-02-22
MOTECH-2941: Fixed MDS period modal

mkruszynski:


kkedys:
@mkruszynski this issue is fixed. But if it is possible please remove the information: "The given value is incorrect. Does not match period format." in Entity schema. Now it appears all the time, even as the format is correct.

kkedys:
@kmadej the issue is fixed.

PR #815, opened 2017-02-27
MOTECH-3138: Fixed multiple handling one task.

pkornowski:


kkedys:
@kmadej the issue is fixed.

PR #817, opened 2017-03-02
MOTECH-3127: Improved showing only the latest task activity by activity type.

pkornowski:


kkedys:
@mkruszynski the issue is resolved.

PR #818, opened 2017-03-03
MOTECH-2929: Fixed remaining empty pages in Web Security tables

mkruszynski:


kkedys:
@pkornowski the issue is fixed.

PR #819, opened 2017-03-06
MOTECH-2742: Fixed support for reset password

mkruszynski:


kkedys:
I verified the ticket. Because the e-mail address is not correct, the following message appears: "No user with the e-mail you entered". When the e-mail is correct, the following message appears:  "An email containing recovery instructions has been sent, check your email." The problem is that I am not capable to verify if that mail arrives on the e-mail. I guess to do that we have to start up the e-mail server.

PR #820, opened 2017-03-07
MOTECH-2657: Fixed prompts in date task fields

mkruszynski:


mkruszynski:
@kkedys This issue was also occurring on Opera browser, not only on Mozilla Firefox.

kkedys:
@pkornowski, the issue is resolved.

PR #821, opened 2017-03-10
MOTECH-3145: Added task action to Query and Update MDS Table

pkornowski:


platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/ActionHandlerServiceImpl.java
```diff
@@ -98,6 +114,19 @@ public void delete(Map<String, Object> parameters) throws ActionHandlerException
         dataService.delete(instance);
     }
 
+    private void setInstancePropertiesForQuery(Object instance, List<Field> fields, Map<String, Object> properties) throws ActionHandlerException {
+        for (Field field : fields) {
+            Object value = properties.get(field.getName());
+            if (value instanceof List) {
```
mkruszynski: You can put it all into one condition and use CollectionUtils.isNotEmpty() on a casted _value_ object.

kkedys:
@pkornowski  When one enters one more time in task, the Date, DateTime and Time disappears.  It also happened before the changes (I verified on CREATE) but because of the Date and DateTime are no longer visible on MDS. Please fix this in the ticket.

kkedys:
@pkornowski Everything works correctly.

PR #822, opened 2017-03-11
MOTEH-3149: Added possibility to manipulate values from bubble with…

pkornowski:
… Map type.


kkedys:
Checked. the value returned is correct, but the caracter size should be the same. Now, in OpenMRS it is: Kate, motech return the value: kate.

kkedys:
checked. The new manipulation works correctly.

PR #823, opened 2017-03-14
MOTECH-3152: Prevented using util Date format as a bubbles.

pkornowski:


kkedys:
@pkornowski Checked. The issue is fixed.

PR #824, opened 2017-03-29
MOTECH-3165: Fixed parsing if a bubble with type map has no data and added UT.

pkornowski:


kkedys:
@pkornowski the issue is fixed.

PR #825, opened 2017-04-04
MOTECH-3153: Fixed parseDate manipulation failures.

pkornowski:


kkedys:
Checked. It works correctly.

PR #826, opened 2017-04-06
Specified concrette dependency verions in the motech-ui.

pkornowski:


kkedys:
Checked. Now the trigger appears and works.

PR #837, opened 2018-04-24
Fixed mds history class loader bug

kmadej:
Steps to reproduce:
1. Create a notification rule instance (it can be other entity with history enabled)
2. Make changes to instance (history rows will be created)
3. Restart the server
4. Go to history of just created notification rule

Expected:
5. History instance should be properly displayed

Actual:
5. Exception is thrown

The problem regards wrong class loader used by datanucleus to retrieve history instance.

To fix this issue, we have to be sure that proper context class loader is set - the same which loads history class

### DIFF
--- a/platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/history/HistoryServiceImpl.java
+++ b/platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/history/HistoryServiceImpl.java
@@ -1,12 +1,12 @@
 package org.motechproject.mds.service.impl.history;
 
 import org.motechproject.mds.domain.EntityType;
-import org.motechproject.mds.service.HistoryTrashClassHelper;
 import org.motechproject.mds.query.Property;
 import org.motechproject.mds.query.PropertyBuilder;
 import org.motechproject.mds.query.QueryParams;
 import org.motechproject.mds.query.QueryUtil;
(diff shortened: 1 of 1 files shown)


platform/mds/mds/src/main/java/org/motechproject/mds/service/impl/history/HistoryServiceImpl.java
```diff
@@ -1,12 +1,12 @@
 package org.motechproject.mds.service.impl.history;
 
 import org.motechproject.mds.domain.EntityType;
-import org.motechproject.mds.service.HistoryTrashClassHelper;
 import org.motechproject.mds.query.Property;
 import org.motechproject.mds.query.PropertyBuilder;
 import org.motechproject.mds.query.QueryParams;
 import org.motechproject.mds.query.QueryUtil;
 import org.motechproject.mds.service.HistoryService;
+import org.motechproject.mds.service.HistoryTrashClassHelper;
```
sheelio: Why is this import moved below HistoryService?

kmadej: Previously imports were in wrong order, after the change org.motechproject.mds.service.HistoryTrashClassHelper is after 
org.motechproject.mds.service.HistoryService, so now we have a proper alphabetical order

